PR #6, opened 2015-03-28
Enabling macros for inspection with edebug.

myeffort:
This addition to macros spec allows edebug to instrument contents of `it` and setup/teardown forms individually. It also captures calculations surrounding them executed on top level of `describe` form.

Instrumented  forms can be invoked in context of the suite later on provided that there will be a command for running known suits or running suite at point without reevaluation of top level form.

Neither currently exists, all I can offer for testing new capabilities is following form, which executes  `buttercup-run-at-point` which instruments suite using edebug before evaluating it:

    (let ((edebug-all-defs t) (edebug-all-forms t)) (buttercup-run-at-point))

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -227,7 +227,7 @@ form.")
 
 (defmacro describe (description &rest body)
   "Describe a suite of tests."
-  (declare (indent 1))
+  (declare (indent 1) (debug (&define sexp def-body)))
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thank you very much!

Usually, I require tests for new features, but for edebug in `declare' statements, I guess the test would be a quite huge for the minimal check it does, so I'll ignore that one here. :-D

PR #17, opened 2015-05-05
Use relative path when testing for dotfiles in current project

### LINKED ISSUE #15: Discovery fails if any parent directory is a dot-directory.
Buttercup helpfully filters tests in dot-directories, so that we won't run tests for packages installed via Cask in `.cask` directory.

However, if I have my project inside `/home/user/.dotdirectory/projects/my-project/tests` where `my-project` is the root of my project, no tests will run because they all match the `/.` of `/.dotdirectory`.

A solution is to cut away the current-directory to only do this filter inside the project directory.

Fuco1:
Fixes #15.


jorgenschaefer:
Thank you very much!

PR #40, opened 2016-02-17
Add assume form to cancel tests if conditions fail

lunaryorn:
The assume form cancels the current test if the given condition is not met.  It is intended to be used in tests that depend on a certain global state that might not be present, e.g. a tool in `exec-path` or a network connection, etc.

Example:

```cl
(describe "Eggs"
  (it "gets eggs from the internet"
    (assume (network-available-p) "No network connection")
    (my/get-eggs)))
```

Output if the assumption is not met

```
Eggs
  gets eggs from the internet  !! CANCELLED !! No network connection
```

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -104,6 +104,20 @@ This is the mechanism underlying `expect'. You can use it
 directly if you want to write your own testing functionality."
   (signal 'buttercup-failed (apply #'format format args)))
 
+(defun buttercup-skip (format &rest args)
+  "Skip the current test with the given description."
+  (signal 'buttercup-pending (apply #'format format args)))
+
+(defmacro assume (condition &optional message)
+  "Assume CONDITIION for the current test.
+
+Assume that CONDITION evaluates to non-nil in the current test.
+If it evaluates to nil cancel the current test with MESSAGE.  If
+MESSAGE is omitted or nil show the condition form instead."
+  (let ((message (or message (format "%S => nil" condition))))
+    `(unless ,condition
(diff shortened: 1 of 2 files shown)


lunaryorn:
Documentation is still missing, but I'm not sure where you'd like to have this form documented.

jorgenschaefer:
Excellent idea, thank you for this!

I'm not sure where to document it, either. I guess there should be a new file in docs/ for non-Jasmine-based extensions.

lunaryorn:
Well, I was approaching buttercup from scalatest and must admit that I found the documentation a little unstructured, to be honest.  In the end I was reading your tests and the source for reference.

lunaryorn:
Specifically I was surprised to find the documentation of forms and matchers in the Introduction whereas Usage only contains installation and setup instructions, and I still make the mistake to click usage when I'd like to see a list of all matchers.

jorgenschaefer:
I can see that confusion, there are two "uses" for Buttercup: Use the library to write tests, and use the package in an environment to run the tests. What would you think a better name for the document that describes how to use buttercup in a project to run tests?

lunaryorn:
@jorgenschaefer Well, as you typically set up your test environment only once (particularly CI like Travis CI), I'd have expected "setup" or so for this document.

But why not name these documents after what they actually describe, i.e. "Run tests", "Write tests", "Setup CI", etc…?

jorgenschaefer:
Hm. I like the names of "writing tests" and "running tests", that's a good idea, thank you!

Issue #9, opened 2015-04-12
Coverage information


jorgenschaefer:
As a buttercup user, I want to know which parts of my code are not covered by my tests in order to identify which parts of my code still need testing.

jorgenschaefer:
This should support non-interactive use, both for setting it up and displaying results, so `testcover.el` is more or less out.

jorgenschaefer:
undercover.el should work just fine, it even ties in with coveralls: https://github.com/sviridov/undercover.el

See sacha's awesome post on the topic: http://sachachua.com/blog/2015/02/continuous-integration-code-coverage-emacs-packages-travis-coveralls/

CodeFalling:
So how to use it with `buttercup`?I try to add
```
(when (require 'undercover nil t)
  (undercover "*.el"))
```
in my every `tests/test-*.el`,but it just fail my test.
This is output from travis-ci:
```
 cask exec buttercup -L .
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/epg-config.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/gnus/mm-decode.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/gnus/ietf-drums.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/help-fns.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/gnus/mm-bodies.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/emacs-lisp/cl-macs.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/gnus/auth-source.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/password-cache.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/gnus/gnus-util.el' newer than byte-compiled file
Source file `/usr/local/evm/emacs-24.4-bin/share/emacs/24.4/lisp/calendar/time-date.el' newer than byte-compiled file
[...]

PR #42, opened 2016-02-21
Added: integration with undercover.el

sviridov:
This PR is related to #9.

It shows how to integrate `emacs-buttercup` with `undercover.el`.

Consider simple case when user has project that uses `emacs-buttercup` for testing. And user want to collect test coverage information for all his top-level `*.el` files (I think it's most common case). To do so user can simply add `(depends-on "undercover")` to `Cask` file. See the [experimenting](https://github.com/sviridov/undercover.el-buttercup-integration-example/tree/experimenting) branch of [undercover.el-buttercup-integration-example](https://github.com/sviridov/undercover.el-buttercup-integration-example/tree/experimenting) as an example.

If user will want to provide extra configuration to `undercover.el` (exclude some files, etc.) user can do it with [UNDERCOVER_CONFIG](https://github.com/sviridov/undercover.el#configuration) environment variable and do something like this:

``` sh
UNDERCOVER_CONFIG='("*.el" (:exclude "awesome-examples.el"))' cask exec buttercup -L .
```

So, what do you think about it?

P.S. While experimenting with `undercover.el-buttercup-integration-example` I faced one strange error when I added `underover` to existed `Cask` file (and `.cask` directory) and did `cask install`:

``` sh
$ cask install
$ TRAVIS=true cask exec buttercup -L . # TRAVIS=true will force undercover to run.

> Invalid function: undercover

$ rm -Rf .cask/
$ cask install
$ TRAVIS=true cask exec buttercup -L .

> Successful output
```

Not sure why that happens.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -659,6 +659,7 @@ current directory."
         (push (car args) dirs)
         (setq args (cdr args)))))
     (setq command-line-args-left nil)
+    (when (require 'undercover nil t) (undercover))
     (dolist (dir (or dirs '(".")))
       (dolist (file (directory-files-recursively
                      dir "\\`test-.*\\.el\\'\\|-test\\.el\\'"))


jorgenschaefer:
Thank you for the contribution, and apologies for the late reply – I apparently postponed the task to check this a week into the future instead of just two days. :-(

I like the idea of adding undercover support, but I do not think it's a good idea to do so automatically. A command line argument like `--undercover` would be nice. I'll take a look at adding such a functionality at some point. The integration is trivial enough!

Thanks for the great package :-)

sviridov:
Hi @bbatsov, well it looks like we should just close this PR.

PR #43, opened 2016-03-17
Debian and Ubuntu installation instructions

spwhitton:
This ELPA package is now available from the Debian and Ubuntu package repositories.


jorgenschaefer:
This is great news, thank you very much for your work!

PR #58, opened 2016-04-10
Add :to-equal-as-set to compare lists as sets.

Fuco1:
Since lists often double as sets in lisp, a primitive to compare them as such is rather useful.

I'm currently writing some "query language parser" to select based on tags (e.g `"foo+bar|baz+!quo"`) and I don't quite care about the order so long as the set returned is correct.

So this simply adds a matcher to do that :P

### DIFF
--- a/buttercup-compat.el
+++ b/buttercup-compat.el
@@ -35,6 +35,9 @@
 (when (not (fboundp 'cl-defstruct))
   (defalias 'cl-defstruct 'defstruct))
 
+(when (not (fboundp 'cl-every))
+  (defalias 'cl-every 'every))
+
 ;;;;;;;;;;;;;;;;;;;;;
 ;; Introduced in 24.4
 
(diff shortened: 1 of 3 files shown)


Fuco1:
Dang, `cl-every` is not avaiable on 24.1/2... hmm.

jorgenschaefer:
You can just add a trivial compatibility definition to `buttercup-compat.el` :-)

jorgenschaefer:
Nice contribution, thank you. I'm a bit unhappy with the name `:to-equal-as-set`. It feels like that describes more or less the implementation, but not the intent. What do you think of `:to-have-same-items` or `:to-have-same-items-as`? [[1]](https://www.npmjs.com/package/jasmine-collection-matchers#expectarray-objecttohavesameitemsarray-object-boolean-ignoresort)

Fuco1:
Fair enough, I'm a mathematician not a writer :D  I'll squash it with the compat commit as well.

Fuco1:
Should be good now.

jorgenschaefer:
Thank you!

PR #60, opened 2016-05-04
feature/windows-bin-support: Added a batch script that execs the butt…

howdoicomputer:
…ercup script using bash

### DIFF
--- a/bin/buttercup.bat
+++ b/bin/buttercup.bat
@@ -1,0 +1,1 @@
+@bash %~dp0buttercup


howdoicomputer:
I'm a Windows user and this allows me to run `cask exec buttercup`.

jorgenschaefer:
Nice, and thank you for the contribution! Could you squash the two commits and add a newline at the end of that file? :-)

howdoicomputer:
@jorgenschaefer There we go! Let me know if anything else is needed. This batch script depends on `bash` being installed but makes no effort to check to see that it is. However, I think it'd be pretty easy for an explorer to figure that out. Will be less of an issue when the Windows Subsystem for Linux rolls out.

jorgenschaefer:
I do not have windows so can not test this, but I think this works better than no batch script, so great. Thanks a lot for the contribution! :-)

howdoicomputer:
@jorgenschaefer That was my line of thought as well. This unblocks me for my projects so I'm more than happy to contribute. Thanks for the merge!

PR #61, opened 2016-05-13
Exclude hidden files and dirs in test root

spwhitton:
The `bin/buttercup` shell script should not try to run tests in hidden
subdirectories of the working directory.  In particular, it should not
run tests found in the `.pc/` directory that quilt(1) generates.


jorgenschaefer:
Good point – thank you for the contribution!

PR #63, opened 2016-07-20
Add projectile integration info to Readme

ebpa:
Example integration for projectile using dir-locals that should save people a bunch of time.

### DIFF
--- a/docs/running-tests.md
+++ b/docs/running-tests.md
@@ -102,6 +102,20 @@ different pattern using the `--pattern` command line argument to the
 You can run this command whichever way you like. Common choices
 include a makefile or shell scripts.
 
+## Projectile
+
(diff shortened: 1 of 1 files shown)


ebpa:
Simply a suggestion.  Feedback welcome.

jorgenschaefer:
Hello, and thanks for the suggestion! I hesitate to add this to the README, as the README should not grow too large. But I have activated the wiki feature on the repo, could you just add it to the wiki? :-) Thank you!

ebpa:
I realized I misspoke in the PR (s/Readme/running-tests.md/).  You probably
noticed.

I will; gladly.  Thanks for enabling that on the repository.  Or perhaps
would another `.md` if running-tests.md would be getting too long?  It's
really nice having the documentation linked to from the landing document.
I'm not sure how robust links into the wiki are given they are technically
in a separate branch.  I can do either.

:+1:

jorgenschaefer:
Oh, actually, I had not noticed! :-D In that case, you are right, keeping it in one spot is a good idea.

Thank you for the contribution!

ebpa:
Thank you!  With buttercup you've made testing in elisp a joy!

PR #68, opened 2016-11-28
Report load errors in buttercup-run-discover

ebpa:
Previously file load errors were uncaught resulting in the core error message being printed, but without any context.  In a project with many test files a syntax error may be nearly untenable.

This change reports the error message with the name of the file as context and then resumes loading test files and running the test suite.


jorgenschaefer:
Nice catch, and thank you very much for the contribution! :-)

ebpa:
It's nothing.  Thank you very much for the package and your work maintaining it!  I use it every day :-)

Issue #10, opened 2015-04-12
Colored output


jorgenschaefer:
The test runner output should be colored, with passing specs/suites in green, failing ones in red, and pending ones in yellow.

It should be possible to disable this using a `--no-color` command line argument.

jorgenschaefer:
From #27:

> - Green for successful tests
> - Red for failing tests
> - Orange for pending ones
> - Standard foreground for skipped ones.

lunaryorn:
@jorgenschaefer I'd like to add that unlike #28 I think this will really become important for Flycheck.  We have about 400 ERT unit tests currently (not counting syntax checker integration tests which add another 100 or so). If we convert all of these to proper specs—which will increase the number of tests somewhat as many large tests should be split into multiple small specs— then we'll definitely need colours to keep the output comprehensible.

ebpa:
Note that not all terminals/outputs support color control characters.  Conditionally enabling color ("auto") according to an inference based on the file descriptor / environment variables / other information can save users from confusion.  Many utilities like grep (which has the option --color=[never|always|auto]) implement this behavior.  The logic in this JavaScript project may be helpful: [chalk/supports-color](https://github.com/chalk/supports-color) (specifically [index.js](https://github.com/chalk/supports-color/blob/master/index.js)).

snogge:
Should this issue be closed now?

jorgenschaefer:
Yep, thank you!

PR #70, opened 2017-03-04
Colorize output

xendk:
As penance for submitting the #10 issue for the third time, I thought I'd take a stab at implementing it.

selection_037

The implementation is was slightly complicated by the fact that the reporter needs to print the text of the spec again on completion, in order to add the color. It could probably be improved much by refactoring it and the standard reporter.

I haven't implemented any detection of color support. I looked at Ecukes for inspiration, and it seems to me that it always uses colors. I've implemented a kill-switch though.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -676,6 +676,9 @@ current directory."
           (error "Option requires argument: %s" (car args)))
         (push (cadr args) patterns)
         (setq args (cddr args)))
+       ((member (car args) '("-c" "--no-color"))
+        (setq buttercup-color nil)
+        (setq args (cdr args)))
        (t
         (push (car args) dirs)
         (setq args (cdr args)))))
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Nice! Thank you for this. I'm afraid I'm kind of busy with moving right now, though, so I'll have to review this at a later point. :-(

buttercup.el
```diff
@@ -929,9 +937,110 @@ Calls either `buttercup-reporter-batch' or
...
+          duration))
+       (when (> failed 0)
+         (error ""))))
+
+    (_
+     ;; Fall through to buttercup-reporter-batch implementation.
+     (buttercup-reporter-batch event arg)))
+  )
+
 (defun buttercup--print (fmt &rest args)
```
jorgenschaefer: Is this a lot of duplicated code? Would it be easier to just add `buttercup-colorize` calls to the existing batch reporter, and check for the color option in that function?

xendk: Well, depends on what you think. That was my initial idea, until I realized that in color mode it needs to re-print the spec name, with color, over the colorless version printed when the spec started. So it prints a carriage return at the beginning of the new string.

So it's not just a matter of adding colors, the strings are quite different in the color/no-color cases.

That's why I started out creating a new reporter, but it could be merged into the standard reporter, thought it would beef it up a bit.

buttercup.el
```diff
@@ -929,9 +937,110 @@ Calls either `buttercup-reporter-batch' or
...
+    (white   . 37))
+  "List of text colors.")
+
+(defun buttercup-colorize (string color)
+  "Format STRING with COLOR."
+  (let ((color-code (cdr (assoc color buttercup-colors))))
+    (format "\u001b[%sm%s\u001b[0m" color-code string)
+    )
+    )
+
```
xendk: :-D I'll fix that..

buttercup.el
jorgenschaefer: Hm. Good point that. Probably not worth merging it, then.

jorgenschaefer:
Looks good to me, thank you for the good work! Could you squash the two commits (just squash locally and force push), I'll be happy to merge then :-)

xendk:
Excellent. Squashed and pushed. :-)

jorgenschaefer:
And merged! My, aren't we efficient! :-D

Thanks a lot for your work!

PR #74, opened 2017-03-31
note dependency on Emacs 24+

dunn:
Required due to the use of pcase, which was added in https://github.com/emacs-mirror/emacs/commit/d02c9bcd096c44b4e3d5e2834c75967b56cdecdd

### DIFF
--- a/buttercup-pkg.el
+++ b/buttercup-pkg.el
@@ -1,2 +1,3 @@
 (define-package "buttercup" "1.5"
+  '((emacs "24"))
   "Behavior-Driven Emacs Lisp Testing")
--- a/buttercup.el
+++ b/buttercup.el
@@ -4,6 +4,7 @@
 
 ;; Version: 1.5
 ;; Author: Jorgen Schaefer <contact@jorgenschaefer.de>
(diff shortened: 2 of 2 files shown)


jorgenschaefer:
Thank you for the contribution! Buttercup requires Emacs 24 for pretty much everything it does, as it heavily relies on lexical binding. Honestly, if anyone out there who is still using Emacs 23 for some confusing reason, and put in the effort to backport package.el to their Emacs 23, and then installed buttercup, I assume they know enough to know why things are breaking left and right. What actual problem does this solve?

dunn:
It's mostly for clarity, I guess.  It lets MELPA know to list 24 as the minimum version.  But it's not very important, so if you'd rather keep things more simple I understand!

jorgenschaefer:
I don't have a strong opinion one way or the other, but considering package.el was not in Emacs before 24, depending on 24 (and not, say, 24.3) seems like a bit of superfluous information :-) Let's add the dependency when buttercup stops working with something newer than 24.1, ok?

dunn:
Sounds good!

PR #76, opened 2017-04-15
Make --with-cleanup public so spies can be used without specs

DamienCassou:
This is useful for people who want to use spies but don't want to use `describe` or `it`.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -563,7 +563,7 @@ KEYWORD can have one of the following values:
 
 (defvar buttercup--cleanup-functions nil)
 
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Fine with me, but out of curiosity, why would you use spies outside of describe/it? :-)

DamienCassou:
> Fine with me, but out of curiosity, why would you use spies outside of describe/it? :-)

some people might want to use buttercup's spies without describe/it, e.g., inside an `ert` test for example. I tried `el-mock`, but it was not as feature complete as buttercup's spies.

PR #78, opened 2017-04-15
[Fix #72] Add buttercup-minor-mode

### LINKED ISSUE #72: Imenu generic expressions
I'm not sure how or where to add this (probably as an optional defun people could put on the elisp hook), but I've found useful to have imenu recognize suites/specs when I have larger test files.

```
(add-to-list 'imenu-generic-expression '("Test Suites" "\\((describe\\_> +\\)\"\\(\\_<.+\\_>\\)\"" 2))
(add-to-list 'imenu-generic-expression '("Spec" "\\((it\\_> +\\)\"\\(\\_<.+\\_>\\)\"" 2))
```

This works for me, right now I just add it to every elisp buffer but some people might not prefer that.  I'm not sure what is the proper convention.

Fuco1:
With buttercup minor mode active the following is activated:

- `describe` and `it` forms are fontified with
  `font-lock-keyword-face`.
- `describe` and `it` forms are available from `imenu` for
  quicker access.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1109,5 +1109,29 @@ failed -- The second value is the description of the expectation
             frame (backtrace-frame n)))
     frame-list))
 
+;;;###autoload
+(define-minor-mode buttercup-minor-mode
+  "Activate buttercup minor mode.
+
+With buttercup minor mode active the following is activated:
+
+- `describe' and `it' forms are fontified with
+  `font-lock-keyword-face'.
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Awesome, thank you!

Fuco1:
Hmm, maybe I should also edit the README or add it somewhere to the docs, right now it is a bit undiscoverable.

PR #81, opened 2017-04-17
Fix :to-have-same-items-as  (fix #80)

### LINKED ISSUE #80: :to-have-same-items-as checks set inclusion, not set equality
The implementation of matcher `:to-have-same-items-as` checks that every element of the first collection is included in the second. This means the matcher is ok to have a larger second collection, provided the first collection is a subset. I would have thought `:to-have-same-items-as` would mean equality, not inclusion. Inclusion could be named instead `:to-be-a-subset-of`.

DamienCassou:
This fixes #80 by adapting `:to-have-same-items-as` implementation to what I think the name suggests.

### DIFF
--- a/buttercup-compat.el
+++ b/buttercup-compat.el
@@ -38,6 +38,9 @@
 (when (not (fboundp 'cl-every))
   (defalias 'cl-every 'every))
 
+(when (not (fboundp 'cl-subsetp))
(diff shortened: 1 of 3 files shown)


jorgenschaefer:
Excellent! Thank you so much!

Issue #85, opened 2017-05-25
Add support for assess


DamienCassou:
I would like to add support for assess in buttercup. For example, a `:buffer` property could  trigger the use of [`assess-with-temp-buffers`](http://homepages.cs.ncl.ac.uk/phillip.lord/lentic/assess-doc.html#sec-3-4). I would also like to see spec bodies be encapsulated inside `assess-with-preserved-buffer-list`. What do you think?

jorgenschaefer:
There was some prior discussion about assess in #37. Basically, I'd be happy to use assess for buttercup instead of what buttercup uses internally. I'm not sure how this would work in reality, though. I have two reservations about it, though. First, I do not want a half solution – either buttercup depends on assess and uses it fully, or it does not depend on assess. Not something in between. And also, I would like buttercup to retain all the current functionality in the same way. I do not know if assess can provide that.

PR #88, opened 2017-06-24
Start getting rid of compile-time warnings

DamienCassou:
I've started cleaning the code to remove compile-time warnings. If you think it's worth it I will keep on doing that (in this PR). This patch changes the number of warnings from 98 to 69. I wanted to do that because I plan on working on #85 and prefer to do that in a clean environment.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -4,6 +4,8 @@
 
 ;; Version: 1.9
 ;; Author: Jorgen Schaefer <contact@jorgenschaefer.de>
+;; Package-Requires: ((emacs "24.3"))
+;; URL: https://github.com/jorgenschaefer/emacs-buttercup
 
 ;; This program is free software; you can redistribute it and/or
 ;; modify it under the terms of the GNU General Public License
@@ -49,7 +51,7 @@
 ;;; wrapper function manipulation
 
 (defun buttercup--enclosed-expr (fun)
-  "Given a zero-arg function, return its unevaluated expression.
+  "Given a zero-arg function FUN, return its unevaluated expression.
 
 The function MUST have one of the following forms:
 
@@ -72,7 +74,7 @@ forms are useful if EXPR is a macro call, in which case the
     (`(lambda nil (quote ,expr) . ,rest) expr)
     (`(lambda nil ,expr . ,(pred identity))
      (error "Function contains multiple expressions: %S" fun))
-        (`(lambda ,(pred identity) . ,(pred identity))
+    (`(lambda ,(pred identity) . ,(pred identity))
      (error "Function has nonempty arglist: %S" fun))
     (_ (error "Not a zero-arg one-expression closure: %S" fun))))
 
@@ -89,7 +91,7 @@ environment)."
         (funcall fun)))
 
 (defun buttercup--wrapper-fun-p (fun)
-  "Returns non-nil if FUN is a zero-arg one-expression function."
+  "Return non-nil if FUN is a zero-arg one-expression function."
   (condition-case nil
       (prog1 t
         (buttercup--enclosed-expr fun))
@@ -120,11 +122,11 @@ a call to `save-match-data', as `format-spec' modifies that."
 
 This macro knows three forms:
 
-\(expect arg :matcher args...)
+\(expect ARG :MATCHER ARGS...)
   Fail the current test iff the matcher does not match these arguments.
   See `buttercup-define-matcher' for more information on matchers.
 
-\(expect (function arg...))
+\(expect (function ARG...))
   Fail the current test iff the function call does not return a true value.
 
 \(expect ARG)
@@ -137,6 +139,10 @@ This macro knows three forms:
       ,@wrapped-args)))
 
 (defun buttercup-expect (arg &optional matcher &rest args)
+  "The function for the `expect' macro.
+
+See the macro documentation for details and the definition of
+ARG, MATCHER and ARGS."
   (cl-assert (cl-every #'buttercup--wrapper-fun-p (cons arg args)) t)
   (if (not matcher)
       (progn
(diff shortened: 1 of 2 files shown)


buttercup-compat.el
```diff
@@ -20,7 +20,7 @@
 
 ;;; Commentary:
 
-;; This file provides compatibility definitions for buttercup. These
+;; This file provides compatibility definitions for buttercup.  These
```
jorgenschaefer: Why do you add double spaces after a period? Does Emacs complain about that somewhere?

buttercup.el
```diff
@@ -35,19 +36,23 @@
...
-(require 'cl)
+(eval-when-compile
+  (require 'cl))
+
 (require 'buttercup-compat)
 
 ;;;;;;;;;;
 ;;; expect
 
+;; defined in buttercup-compat
```
jorgenschaefer: What's the purpose of this comment?

buttercup.el
```diff
@@ -145,91 +157,91 @@ MATCHER is either a matcher defined with
...
-  (let ((result (buttercup--apply-matcher matcher (cons obj args))))
-    (if (consp result)
-        (cons (not (car result))
-              (cdr result))
-      (not result))))
+                          (let ((result (buttercup--apply-matcher matcher (cons obj args))))
+                            (if (consp result)
+                                (cons (not (car result))
+                                      (cdr result))
+                              (not result))))
```
jorgenschaefer: Please load buttercup before re-indenting the file, buttercup defines correct indentation rules for `buttercup-define-matcher`.

buttercup.el
```diff
@@ -267,7 +279,7 @@ MATCHER is either a matcher defined with
   failure-stack)
 
 (defun buttercup-suite-add-child (parent child)
-  "Add a CHILD suite to a PARENT suite."
+  "Add to PARENT suite a CHILD suite."
```
jorgenschaefer: I know the Emacs linter requires args in docstrings to be of the same order as in the definition, but this is just extremely awkward. I don't think this is a good docstring.

jorgenschaefer:
You brave soul, trying to make the elisp linter happy … I find quite a few of its rules to be annoying at best, sadly. I'm fine with trying to reduce the number of actual byte code compiler warnings. I'm also fine with trying to make the linter less unhappy, but I'm not fine with making the code or docstrings less readable because of that. So, please keep doing this, but don't expect the end result to be zero warnings, especially from `checkdoc` …

buttercup.el
DamienCassou: none any more. Previously, it was to justify the compile-time warning.

buttercup.el
DamienCassou: sorry about that, I shouldn't have committed.

buttercup.el
DamienCassou: You are right.

DamienCassou:
What do you think about this one? This gets rid of 60 compile-time warnings out of 97.

buttercup.el
```diff
@@ -814,39 +895,13 @@ Do not change the global value.")
 ;;;;;;;;;;;;;
 ;;; Reporters
 
-(defvar buttercup-reporter #'buttercup-reporter-adaptive
```
jorgenschaefer: I'm somewhat unhappy with moving this variable out of the "Reporters" section. It belongs here. Sadly, the block can not be moved as a whole, because it itself needs previous code. Elisp should be smarter about "undefined variable" references, really. (Good example where stupid linter rules reduce clarity, instead of increasing it.)

tests/test-buttercup.el
```diff
@@ -15,6 +15,9 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+;;; Commentary:
+;;
+
```
jorgenschaefer: Adding an empty commentary because some linter requires it does not improve readability, either.

tests/test-buttercup.el
```diff
@@ -266,7 +269,7 @@
               "su1 su2"))))
 
 (describe "The `buttercup-spec-full-name' function"
-  (let (su1 su2 sp1)
+  (let (su1 su2 sp1 sp2)
```
jorgenschaefer: On a brighter note, this is what linters are there for: Catching bugs and improving readability! Nice. :-)

buttercup-compat.el
DamienCassou: Yes. The variable `sentence-end-double-space` is set to true by default.

buttercup.el
DamienCassou: What about letting the variable in the "Reporters" section and adding

```
(eval-when-compile
  ;; Defined below in a dedicated section
  (defvar buttercup-reporter))
```

before its first use?

tests/test-buttercup.el
DamienCassou: Do me this is an example of a case where I don't care about the empty section but I care about the absence of warnings in my file. I added some comments.

tests/test-buttercup.el
DamienCassou: In my opinion, this can only work if there is 0 warnings on the file. If the file is full of warnings I should accept, I will just stop having a look at them and miss the important ones. That's why I sometimes accept a few workarounds to shut up the stupid rules of the linter. A better approach would be to disable some rules either per line or per section or per file (as with eslint for example).

DamienCassou:
What about merging this PR now?

buttercup-compat.el
jorgenschaefer: Then people working on buttercup should change the default. It's a silly one.

buttercup.el
```diff
@@ -4,6 +4,7 @@
 
 ;; Version: 1.7
 ;; Author: Jorgen Schaefer <contact@jorgenschaefer.de>
+;; Package-Requires: ((emacs "24.4"))
```
jorgenschaefer: This is incorrect. buttercup works with Emacs 24.3 (and even older versions, really, just not tested).

buttercup.el
```diff
@@ -580,10 +628,10 @@ KEYWORD can have one of the following values:
   (gethash (symbol-function spy)
            buttercup--spy-contexts))
 
-(defun buttercup--spy-calls-add (spy-function context)
+(defun buttercup--spy-calls-add (spy context)
```
jorgenschaefer: In other functions, `spy` refers to the symbol, not the function value. Here, it's the function value. Admittedly, the naming is not consistent, but I'd rather replace all occurrences of `spy` with the more explicit version than to conflate them even more.

tests/test-buttercup.el
```diff
@@ -824,3 +830,10 @@
         (expect specs :to-be-truthy)
         (expect (length (cdr specs)) :to-equal 1)
         (expect (caadr specs) :to-equal "should fontify special keywords")))))
+
+(provide 'test-buttercup)
+;;; test-buttercup.el ends here
+
+;; Local Variables:
+;; flycheck-emacs-lisp-load-path: ("..")
+;; End:
```
jorgenschaefer: The Local Variables section should not be here.

buttercup.el
DamienCassou: lexical-binding, as declared on top of the file, requires at least Emacs 24. `autoloadp`, as used in `spy-on`, requires at least 24.3. I fixed it to 24.3. Do you want to add `autoloadp` to compat?

buttercup-compat.el
DamienCassou: done

jorgenschaefer:
This looks good to me now – thank you!

PR #91, opened 2017-06-30
Implement :to-have-been-called-times matcher (close #90)

### LINKED ISSUE #90: :to-have-been-called-times
I would like to have something like that (untested):

```emacs
(buttercup-define-matcher :to-have-been-called-times (spy number)
  (let* ((call-count (length (spy-calls-all spy))))
    (cond
     ((= number call-count)
      t)
     (t
      (cons nil
            (format "Expected `%s' to have been called %s %s, but it was called %s %s"
                    spy
                    number (if (= number 1) "time" "times")
                    call-count (if (= call-count 1) "time" "times")))))))
```

do you want a (tested) PR?

DamienCassou:


DamienCassou:
I'm going on vacation now and won't be really reactive for the next 3 weeks. I will still read my emails regularly though.

jorgenschaefer:
Thank you for your work!

PR #92, opened 2017-06-30
Set Emacs 24.3 as minimum and clean outdated code (close #89)

### LINKED ISSUE #89: Compatibility with old versions of Emacs
Why do you keep compatibility with very old versions of Emacs (Emacs 24.1 is from 2012)? Buttercup is meant for Emacs package developers who are probably using recent versions of Emacs. If you care about Emacs developers using old Emacs versions, what about keeping compatibility with Emacs 24.5 instead of 24.1? Increasing minimum Emacs version would simplify the code base (e.g., cleaning of `-compat.el` and use of `cl-lib`) and reduce compile-time warnings (use of `define-error`).

DamienCassou:


jorgenschaefer:
Excellent, thank you!

PR #93, opened 2017-07-21
Print unevaluated expression when reporting a failure

DarwinAwardWinner:
The `expect` macro now passes `args` as a single list to `buttercup-expect` (instead of as the `&rest` arguments), and then passes a 4th argument which is the unevaluated expression that was passed to `expect`. `buttercup-expect` uses this to add a prefix to the failure message containing the unevaluated expression. This solution is kind of clunky, but requires no changes to existing matchers. With this change, now `(expect (+ 1 1) :to-equal 3)` fails with ``While evaluating `(+ 1 1)': Expected 2 to `equal' 3``.

The tests for the macro expansion of `expect` have been modified accordingly.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -246,7 +246,8 @@ MATCHER is either a matcher defined with
                                  "of %S, but it threw %S")
                          function signal (car err)))))
       (t
-       (cons t (format "Expected %S not to throw an error" function)))))))
+       (cons t (format "Expected %S not to throw an error, but it threw %S"
+                       function err)))))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Suite and spec data structures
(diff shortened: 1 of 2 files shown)


DarwinAwardWinner:
This could probably be made less clunky by converting `buttercup-expect` to take keyword arguments using `cl-defun`. Let me know if you're interested in that.

DarwinAwardWinner:
For extra long expressions, this would probably benefit from formatting the expression over multiple lines using `pp-to-string`.

DarwinAwardWinner:
Another minor issue: expressions inside functions (e.g. for the `:to-throw` matcher) will be macro-expanded in the failure message, unless the lambda form is quoted (with `'`, not `#'`).

DarwinAwardWinner:
By the way, don't merge this branch, because #100 makes it obsolete.

PR #95, opened 2017-07-28
Reimplement "spy-on" to work with commands

DarwinAwardWinner:
The new implementation preserves the interactive form of its argument,
and warns if the replacement passed to :and-call-fake changes the
interactive form. This ensures that spied commands can still be called
as commands (i.e. via "command-execute").

I haven't added any tests for this yet.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -550,28 +550,54 @@ KEYWORD can have one of the following values:
       the original function.
 
   nil -- Track calls, but simply return nil instead of calling
-      the original function."
-  (cond
-   ((eq keyword :and-call-through)
-    (let ((orig (symbol-function symbol)))
-      (buttercup--spy-on-and-call-fake symbol
-                                       (lambda (&rest args)
-                                         (apply orig args)))))
-   ((eq keyword :and-return-value)
-    (buttercup--spy-on-and-call-fake symbol
-                                     (lambda (&rest args)
-                                       arg)))
-   ((eq keyword :and-call-fake)
-    (buttercup--spy-on-and-call-fake symbol
-                                     arg))
-   ((eq keyword :and-throw-error)
-    (buttercup--spy-on-and-call-fake symbol
-                                     (lambda (&rest args)
-                                       (signal arg "Stubbed error"))))
-   (t
-    (buttercup--spy-on-and-call-fake symbol
-                                     (lambda (&rest args)
-                                       nil)))))
+      the original function.
+
+If the original function was a command, the generated spy will
+also be a command with the same interactive form, unless
+`:and-call-fake' is used, in which case it is the caller's
+responsibility to ensure ARG is a command."
+  ;; We need to load an autoloaded function before spying on it
+  (when (autoloadp (symbol-function symbol))
+    (autoload-do-load (symbol-function symbol) symbol))
+  (cl-assert (not (autoloadp (symbol-function symbol))))
+  (let* ((orig (symbol-function symbol))
+         (orig-intform (interactive-form orig))
+         (replacement
+          (pcase
+              keyword
+            (:and-call-through
+             (when arg
+               (error "`spy-on' with `:and-call-through' does not take an ARG"))
+             `(lambda (&rest args)
+                ,orig-intform
+                (apply ',orig args)))
+            (:and-return-value
+             `(lambda (&rest args)
+                ,orig-intform
+                ,arg))
+            (:and-call-fake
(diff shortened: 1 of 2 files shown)


buttercup.el
```diff
@@ -560,7 +579,12 @@ KEYWORD can have one of the following values:
                                  :return-value return-value
                                  :current-buffer (current-buffer)))
               return-value)))
-    this-spy-function))
+    ;; Add the interactive form from `fake-function', if any
+    (if (interactive-form fake-function)
+        `(lambda (&rest args)
+           ,(interactive-form fake-function)
+           (apply ',this-spy-function args))
```
jorgenschaefer: I think this can be `(,this-spy-function ,@args)`, no?

tests/test-buttercup.el
```diff
@@ -501,17 +501,40 @@
 
 (describe "The Spy "
   (let (test-function)
-    (before-each
+    (before-all
```
jorgenschaefer: Why this change? If one of the tests breaks the binding, all tests would fail now – that's not a particularly useful co-dependence. And the setup is not really expensive.

buttercup.el
```diff
@@ -518,28 +518,47 @@ KEYWORD can have one of the following values:
...
-                                     (lambda (&rest args)
-                                       nil)))))
+      the original function.
+
+If the original function was a command, the generated spy will
+also be a command with the same interactive form, unless
+`:and-call-fake' is used, in which case it is the caller's
+responsibility to ensure ARG is a command."
+  (let* ((orig (symbol-function symbol))
+        (orig-intform (interactive-form orig))
```
jorgenschaefer: I think the indentation is wrong here, but that might be github's display.

buttercup.el
```diff
@@ -518,28 +518,47 @@ KEYWORD can have one of the following values:
...
+        (replacement
+         (cl-case
+             keyword
```
jorgenschaefer: I do not have a strong opinion here, but I think the idiomatic elisp macro would be `pcase` … not sure either of them really add much readability over the cond form.

jorgenschaefer:
Very nice, thank you for your work! This is great!

buttercup.el
DarwinAwardWinner: I probably forgot to reindent after switching `let` to `let*`. I'll fix it.

tests/test-buttercup.el
DarwinAwardWinner: It's explained in the commit message: https://github.com/jorgenschaefer/emacs-buttercup/pull/95/commits/653f7edc2f431a0ebb983a8d3624ad7c24b9adae

Basically, some of the tests aren't actually testing anything if you use `before-each`. For example, the "restores the old value after a spec run" will always pass because the `before-each` will restore the old value even if `spy-on` doesn't.

jorgenschaefer: Interesting. I tricked myself with the semantics there :-D

I think it'd be good to add a comment regarding this here, to prevent others from assuming a quick fix there.

buttercup.el
DarwinAwardWinner: No, because `this-spy-function` is a variable containing a function (i.e. we need `(symbol-value 'this-spy-function)`. And `args` doesn't need un-quoting.

jorgenschaefer: My bad, misread the code. :-)

DarwinAwardWinner:
I just added a failing test: after running a spied-on command via `command-execute`, the spied-on command doesn't match `:to-have-been-called`.

DarwinAwardWinner:
Actually, #103 could probably be solved completely and cleanly by reimplementing `spy-on` using `advice-add` and `advice-remove` from nadvice.el, which know how to handle commands, autoloads, etc. The problem is that nadvice was only added in Emacs 24.4, so you'd have to drop support for 24.3, which is currently included in the list of versions tested on Travis.

I suppose the old advice system could do the job as well, but it would be more difficult. 

In any case, I've squashed this into two commits, one with the fix to the spy test logic, and one with the new, improved `spy-on` that checks for the right number of arguments and works on commands and autoloads. If you're interested in pursuing the advice-based approach, let me know.

jorgenschaefer:
Using nadvice might be a good idea, but as long as package authors might want to use buttercup to test packages for 24.3, that's sadly not an option.

Also some conflicts in this PR I'm afraid :-(

DarwinAwardWinner:
Ok, I resolved the conflicts. It was a trivial conflict, just two `require` lines added in the same place.

PR #96, opened 2017-07-29
Ensure elc files are up to date before running tests

DarwinAwardWinner:
The test rule in the Makefile now depends on the compile rule, so
there should be no more warnings about out-of-date elc files when
running tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,20 @@
 EMACS := emacs
-VERSION = $(shell sed -ne 's/^;; Version: \(.*\)/\1/p' buttercup.el)
-DISTFILES = buttercup.el buttercup-compat.el buttercup-pkg.el README.md
+VERSION := $(shell sed -ne 's/^;; Version: \(.*\)/\1/p' buttercup.el)
+ELISP_FILES := $(shell ls *.el | grep -v -- '-pkg\.el$$')
+DISTFILES := $(ELISP_FILES) buttercup-pkg.el README.md
 
-
-.PHONY: test
+.PHONY: test compile clean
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Hm. This also means you can not run tests without compiling. I guess that's fine.

But it also means you have to change the makefile if we ever add more files. I don't like that too much. Wasn't there some kind of globbing for makefiles for this?

DarwinAwardWinner:
I added wildcard-based globbing to the Makefile. The disadvantage (which also applies to the original Makefile) is that it also compiles the `buttercup-pkg.el` file, which it technically shouldn't.

jorgenschaefer:
Ah, true. Do you think it's worthwhile to use filter-out? Or just revert to your original code? Or get rid of the makefile and just use shell scripts?

DarwinAwardWinner:
I switch to shell globbing and then `grep -v` to get rid of the package file. I also DRY'd it with the list of files at the top of the Makefile.

jorgenschaefer:
Your Makefile-fu is definitely above mine :-D Nice work, thank you! Would you like to squash and rebase, or should I let github handle this?

DarwinAwardWinner:
Ok, squashed into one commit and rebased onto latest master.

DarwinAwardWinner:
(As for my Makefile-fu, it comes from a few years ago when I tried way to hard to get GNU Make to work as a workflow executor for a complex bioinformatics workflow.)

jorgenschaefer:
Thank you very much for your contribution! :-)

PR #97, opened 2017-07-29
Signal an error when ":to-throw" is used on a non-function

### LINKED ISSUE #53: :to-throw doesn't check if its argument is a function
The following test passes:

``` elisp
(describe "Buttercup"
  (it "does not complain when the argument of :to-throw isn't a function"
    (expect nil :to-throw 'error)))
```

That's clearly a user error, but it would be nice to get a warning about it: I got bitten by this because my function didn't throw, and I set the test up incorrectly (as in #41), so I thought everything was fine.

DarwinAwardWinner:
Comes with a test.

Fixes #53.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -203,6 +203,10 @@ MATCHER is either a matcher defined with
                       a b precision))))
 
 (buttercup-define-matcher :to-throw (function &optional signal signal-args)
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Very nick quick win solution to that problem, thank you! :-D

PR #98, opened 2017-07-29
Show a message about duplicate spec names

### LINKED ISSUE #77: Detect same spec names
Jasmine and buttercup both accepts similarly named specs:

```emacs
(describe "foo"
  (it "bar" ...)
  (it "bar" ...))
```

But, whereas Jasmine has [no-spec-dupes](https://github.com/tlvince/eslint-plugin-jasmine/blob/master/docs/rules/no-spec-dupes.md) to detect duplicate specs, I don't know anything related for buttercup.

DarwinAwardWinner:
This should be a warning, but for some reason "display-warning" is a
no-op inside of "buttercup-describe". If you know how to fix this, please 
let me know.

Fixes #77.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -351,6 +351,23 @@ MATCHER is either a matcher defined with
                 (buttercup-spec-description spec))
       (buttercup-spec-description spec))))
 
+(defun buttercup--full-spec-names (spec-or-suite-list)
+  "Return full names of all specs in SPEC-OR-SUITE-LIST."
+  (cl-loop
+   for x in (buttercup--specs-and-suites spec-or-suite-list)
+   if (buttercup-spec-p x)
+   collect (buttercup-spec-full-name x)))
+
+(defun buttercup--find-duplicate-spec-names (spec-or-suite-list)
+  "Return duplicate full spec names among SPEC-OR-SUITE-LIST."
+  (let ((seen '())
+        (duplicates '()))
+    (dolist (name (buttercup--full-spec-names spec-or-suite-list)
+                  (nreverse duplicates))
+      (if (member name seen)
+          (push name duplicates)
+        (push name seen)))))
+
 ;;;;;;;;;;;;;;;;;;;;
 ;;; Suites: describe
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Nice idea, thank you.

Your elisp has a strong CL accent. :-D I am generally not too opposed to using `cl` stuff, but that second loop is getting quite a bit advanced. I'll have to think about whether I want that kind of code. (I am aware that the alternatives are less readable, which is why I am considering it at all.)

DarwinAwardWinner:
I'm happy to rewrite it using whatever style you want. I'm a strong proponent of getting it working first and editing for style later.

DarwinAwardWinner:
The irony is I've never actually programmed in CL. I just like the cl libraries in elisp.

jorgenschaefer:
> The irony is I've never actually programmed in CL. I just like the cl libraries in elisp.

Hah! Nice :-) I never got warm with it, for various reasons.

Having thought about the loop use there, I'm fine with it (due to readability), but I would prefer if that code was in a separate function, `buttercup--find-duplicates` or something. Makes it easier to not get confused by the presence of non-idiomatic code in the middle of an unrelated function.

DarwinAwardWinner:
Ok, I did the refactoring, and rewrote it using `dolist`.

Oh, one problem with this is that apparently warnings don't work from wherever this code is called from, which is why I had to resort to `message` instead. I tried using `display-warning` and it did nothing. Do you have any ideas about that?

jorgenschaefer:
The code looks good to me. Would you like to squash and rebase yourself, or should I let github handle that?

As for the `display-warning` issue, I have no idea. The whole warnings facility is rather baroque, and I do not quite understand when things are printed and when not. Possibly a combination of interactive uses, the `*Messages*` buffer, and the like?

DarwinAwardWinner:
This branch is fully squashed and ready to merge, by the way.

jorgenschaefer:
Thank you! :-)

PR #99, opened 2017-07-30
Configurable stack traces

DarwinAwardWinner:
This implements two variables to control the printing of stack traces. `buttercup-stack-frame-style` can be set to `short`, `long`, or `pretty`. `buttercup-stack-frame-add-prefixes` can be set non-nil to prefix each frame with a lambda for function calls or an M for macros and special forms.

I still need to add a command-line argument for this.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -716,6 +716,14 @@ current directory."
         (args command-line-args-left))
     (while args
       (cond
+       ((member (car args) '("--traceback"))
+        (when (not (cdr args))
+          (error "Option requires argument: %s" (car args)))
+        ;; Make sure it's a valid style by trying to format a dummy
+        ;; frame with it
+        (buttercup--format-stack-frame '(t myfun 1 2) (intern (cadr args)))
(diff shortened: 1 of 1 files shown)


DarwinAwardWinner:
I added a command-line argument to control trace style.

DarwinAwardWinner:
FYI If you don't like the lambda/M prefixes, you can just skip 82e7e58.

jorgenschaefer:
After thinking about it for a moment, I like the prefixes – but I don't think there should be an extra option for them, they can be the default for "pretty" stack traces and not there for the rest? What do you think?

DarwinAwardWinner:
Sounds good to me. I'll make that change.

DarwinAwardWinner:
Ok, I implemented that change and squashed everything.

jorgenschaefer:
This seems to conflict with one of your other changes I'm afraid, could you look again?

DarwinAwardWinner:
Ok, I rebased it onto master.

Issue #41, opened 2016-02-19
Wrap arguments of expect in closures to allow :to-throw calls to be more intuitive


snogge:
I tried to write tests to verify errors on bad arguments and was bitten by the fact that

```
(expect (tested-function 'foo) :to-throw)
```

should have been 

```
(expect (apply-partially 'tested-function 'foo) :to-throw)
```

A quick look at the expect macro shows that this is probably the easiest solution, but it feels backward and unintuitive.

jorgenschaefer:
Hello, thanks for the suggestion! This is not quite as obvious as it might seem. Depending on what the programmer does, Buttercup might have to evaluate the argument first before getting a function, or not. E.g. `(expect (car function-list) :to-throw)` would require the argument to evaluate first. And once you do not expand the arguments right away, you get strange behavior with scoping. So, `expect` simply does evaluate everything before passing it to the matchers. That should be simple to remember.

Incidentally, this works, too: `(expect (lambda () (tested-function 'foo)) :to-throw)`

Does this help with the confusion? :-)

snogge:
I understand why, but was first surprised by the fact that I needed to pass a function and so needed different syntax specifically for the `:to-throw` matcher. 

A solution could be to always catch signals in the expect macro, and send the signal to the matcher instead of the function object or other return value. This would probably make error messages of other matchers confusing when unexpected signals are thrown. 

If the matcher can be marked as expecting signals and signals only caught for marked matchers the non-catching matchers could behave as before. 

Anyway, my point is that expect declarations using `:to-throw` have to be written differently, which makes the interface less elegant.

jorgenschaefer:
Hm. Come to think of it, the `expect` macro could wrap all arguments in closures and pass them to the `buttercup-expect` function, which would enable this feature. I think I like that idea, thank you for bringing up that this is confusing and also for being patient with me in explaining it! :-)

snogge:
That does sound like a better solution. Looking forward too seeing it in 1.5 :)

jorgenschaefer:
Hm. I just tried to implement this – it's sadly rather non-trivial. Basically, every single matcher would have to be adjusted to incorporate this and the default error display would break, or matchers would need some (hacky) plist entry to mark them as "delayed execution" … I dislike the latter greatly. This requires more thought, it seems. Maybe in combination with #29?

DarwinAwardWinner:
[...]

Issue #29, opened 2016-01-10
Syntax/function to create return values for matchers


jorgenschaefer:
Currently, the return value of matchers is pretty unreable. There should be a function or even syntax to create the return value that makes it obvious what is happening.

See #21 for some discussion.

jorgenschaefer:
Something like this, maybe:

```Lisp
(buttercup-define-matcher :to-be-local (symbol)
  (buttercup-matches (local-variable-p symbol)
    :if-false (format "Expected %S to be a local variable, but it is not"
                      symbol)
    :if-negated (format "Expected %S not to be a local variable, but it is"
                        symbol)))
```

DarwinAwardWinner:
I've implemented this as part of my expect-closures branch here: https://github.com/jorgenschaefer/emacs-buttercup/pull/100/files#diff-9ec0e003243d9ae887c64b1600f7c32aR183

DarwinAwardWinner:
I think this is solved by the `buttercup--test-expectation` macro.

PR #100, opened 2017-07-30
Closure-based "expect" macro, fully backward-compatible

DarwinAwardWinner:
Ok, here's the big one. As discussed in #41, I expanded my proof of concept into a fully working prototype. The `expect` macro now wraps all its arguments in closures before passing them to `buttercup-closure-expect` (the replacement for `buttercup-expect`). This means that closure-based matchers can now see the un-evaluated expressions that were passed to `expect` and include them in failure messages. I have defined a minimal set of such closure-based matchers, namely `::to-be`, `::to-equal`, `::not`, and `::to-be-non-nil` (all with double colon prefixes to avoid shadowing existing matchers). However, the closure-based `expect` will transparently fall back to standard non-closure-based matchers (i.e. anything defined by `buttercup-define-matcher`), so it is completely backward-compatible with the existing syntax, and we can gradually migrate matchers one by one to the new closure-based system, and the only thing users should notice is better error messages.

The first 3 commits in this pull request implement some minor refactoring to simplify the expect/matcher logic and make it easier to translate that logic to closures. The 4th commit actually implements the closure-based `expect`, and it does so in a separate file, so just loading buttercup.el will still give you the old `expect`. Loading the new file will redefine `expect` to be the new macro. I don't expect this arrangement to stay like this, but for now it makes it easy to turn the closure mode on and off for testing.

The relevant tests have been modified for the new code, since obviously the macro expansion of `expect` is now different. New tests should probably be added for the new functions in buttercup-closure-matchers.el.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1218,13 +1218,13 @@ current directory."
     (dolist (file command-line-args-left)
       (with-current-buffer (find-file-noselect file)
         (goto-char (point-min))
-        (while (let ((case-fold-search t))
-                 (re-search-forward
+        (let ((case-fold-search t))
+          (while (re-search-forward
                   "```\\(?:emacs-\\|e\\)?lisp\n\\(\\(?:.\\|\n\\)*?\\)```"
-                  nil t))
-          (let ((code (match-string 1)))
-            (with-current-buffer lisp-buffer
-              (insert code))))))
+                  nil t)
+            (let ((code (match-string 1)))
+              (with-current-buffer lisp-buffer
+                (insert code)))))))
     (with-current-buffer lisp-buffer
       (setq lexical-binding t)
       (eval-region (point-min)
--- a/buttercup.el
+++ b/buttercup.el
@@ -800,8 +800,10 @@ current directory."
     (dolist (file command-line-args-left)
       (with-current-buffer (find-file-noselect file)
         (goto-char (point-min))
-        (while (re-search-forward "```lisp\n\\(\\(?:.\\|\n\\)*?\\)```"
-                                  nil t)
+        (while (let ((case-fold-search t))
+                 (re-search-forward
+                  "```\\(?:emacs-\\|e\\)?lisp\n\\(\\(?:.\\|\n\\)*?\\)```"
+                  nil t))
           (let ((code (match-string 1)))
             (with-current-buffer lisp-buffer
               (insert code))))))
--- a/buttercup.el
+++ b/buttercup.el
@@ -772,10 +772,7 @@ current directory."
       (dolist (file (directory-files-recursively
                      dir "\\`test-.*\\.el\\'\\|-tests?\\.el\\'"))
         (when (not (string-match "\\(^\\|/\\)\\." (file-relative-name file)))
-          (condition-case err
-              (load file nil t)
-            (error
-             (message (format "Error while loading %s:\n%s" file (error-message-string err))))))))
+          (load file nil t))))
     (when patterns
       (let ((suites-or-specs buttercup-suites))
         (while suites-or-specs
--- a/buttercup.el
+++ b/buttercup.el
@@ -45,51 +45,54 @@
 (require 'ert nil t)
 (require 'warnings)
 
-;;;;;;;;;;;;;;;;;;;;;;;;
-;;; closure manipulation
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; wrapper function manipulation
 
-(defun buttercup--enclosed-expr (x)
-  "Given a zero-arg closure, return its unevaluated expression.
+(defun buttercup--enclosed-expr (fun)
+  "Given a zero-arg function, return its unevaluated expression.
 
-The closure MUST have one of the following forms:
+The function MUST have one of the following forms:
 
+\(lambda () EXPR)
 \(closure (ENVLIST) () EXPR)
+\(lambda () (quote EXPR) EXPR)
 \(closure (ENVLIST) () (quote EXPR) EXPR)
 
-and the return value will be EXPR, unevaluated. The latter form
-is useful if EXPR is a macro call, in which case the `quote'
-ensures access to the un-expanded form."
-  (pcase x
+and the return value will be EXPR, unevaluated. The latter 2
+forms are useful if EXPR is a macro call, in which case the
+`quote' ensures access to the un-expanded form."
+  (pcase fun
     (`(closure ,(pred listp) nil ,expr) expr)
     (`(closure ,(pred listp) nil (quote ,expr) . ,rest) expr)
     (`(closure ,(pred listp) nil ,expr . ,(pred identity))
-     (error "Closure contains multiple expressions: %S" x))
+     (error "Closure contains multiple expressions: %S" fun))
     (`(closure ,(pred listp) ,(pred identity) . ,(pred identity))
-     (error "Closure has nonempty arglist: %S" x))
+     (error "Closure has nonempty arglist: %S" fun))
     (`(lambda nil ,expr) expr)
     (`(lambda nil (quote ,expr) . ,rest) expr)
     (`(lambda nil ,expr . ,(pred identity))
-     (error "Function contains multiple expressions: %S" x))
+     (error "Function contains multiple expressions: %S" fun))
         (`(lambda ,(pred identity) . ,(pred identity))
-     (error "Function has nonempty arglist: %S" x))
-    (_ (error "Not a zero-arg one-expression closure: %S" x))))
+     (error "Function has nonempty arglist: %S" fun))
+    (_ (error "Not a zero-arg one-expression closure: %S" fun))))
 
-(defun buttercup--closure-expr-and-value (x)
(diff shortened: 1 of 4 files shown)


DarwinAwardWinner:
Just to give a taste, here's the old system:

```elisp
(pp
 (list
 (condition-case err
     (expect (functionp t) :not :to-be nil)
   (error err))
 (condition-case err
     (expect (functionp 'ignore) :to-be nil)
   (error err))))
```

which prints:

```
((buttercup-failed . "Expected nil not to be `eq' to nil")
 (buttercup-failed . "Expected t to be `eq' to nil"))
```

And here's the same test with the closure matchers:

```elisp
(pp
 (list
  (condition-case err
      (expect (functionp t) ::not ::to-be nil)
    (error err))
  (condition-case err
      (expect (functionp 'ignore) ::to-be nil)
    (error err))))
```

which gives

```
((buttercup-failed . "Expected (functionp t) not to be `eq' to nil")
 (buttercup-failed . "Expected (functionp (quote ignore)) to be `eq' to nil but instead it was t"))
```

DarwinAwardWinner:
Also, I like to compliment the quality of your coding, since I've submitted like six simultaneous orthogonal pull requests, and they all merge mostly cleanly.

DarwinAwardWinner:
So, I just realized that the `:to-be-truthy` matcher is equivalent to `:to-be-non-nil`. I reimplemented it because I had assumed that it was a wrapper for the [truthy](https://github.com/rolandwalker/truthy), which tests for nonempty content rather than non-nil value.

DarwinAwardWinner:
Full disclosure, this is effectively an implementation of what you said [here](https://github.com/jorgenschaefer/emacs-buttercup/issues/41#issuecomment-192964036):

> matchers would need some (hacky) plist entry to mark them as "delayed execution" … I dislike the latter greatly

since I'm just stroing the closure-based matcher in a different plist entry. But I hope I've made the interface to the whole thing intuitive enough to make you hate it less.

jorgenschaefer:
This is really nice, thank you very much! This should be the default in buttercup.el, with the old code being removed (and single-colon matchers used :-)). Do you think you could just do that?

DarwinAwardWinner:
Ok, I'll replace the old matchers with the new ones, and incorporate the new code into buttercup.el rather than leaving it in a separate file. There's a couple more important questions on how to fully integrate this, though:

- Are you fine with the naming scheme? I basically added `closure` to all the function and variable names to disambiguate them from their still-existing non-closure counterparts. Should I remove that, since the old functions are mostly going away?
- Should `buttercup-define-matcher` remain for the sake of backward compatibility, in case any external packages were using it to define their own matchers? Should it be marked as deprecated?
- The `:to-throw` matcher, which previously required a lambda form, now only requires an expression. Passing a lambda form will result in the function not being called. Should the matcher check for a lambda form and issue a warning, to help people migrate to the new usage?
- Do you like the convenience functions and macros I've defined? Namely `buttercup--test-expectation`, `buttercup-define-closure-matcher-for-unary-function`, and `buttercup-define-closure-matcher-for-binary-function`. (See also #29)

Those are the ones I can think of right no. There might be other questions that come up as I do the integration. Also, a lot of docstrings need updating, which I will do my best on as well.

jorgenschaefer:
> Are you fine with the naming scheme? I basically added closure to all the function and variable names to disambiguate them from their still-existing non-closure counterparts. Should I remove that, since the old functions are mostly going away?

Yes, please remove that :-)

> Should buttercup-define-matcher remain for the sake of backward compatibility, in case any external packages were using it to define their own matchers? Should it be marked as deprecated?

I'd just replace it. We can bump Buttercup's major version number for this to make it a bit more obvious that we're breaking things.

> The :to-throw matcher, which previously required a lambda form, now only requires an expression. Passing a lambda form will result in the function not being called. Should the matcher check for a lambda form and issue a warning, to help people migrate to the new usage?

Excellent idea! (For the time being, the function created by the macro can check if the only value is a function and warn)

> Do you like the convenience functions and macros I've defined? Namely buttercup--test-expectation, buttercup-define-closure-matcher-for-unary-function, and buttercup-define-closure-matcher-for-binary-function. (See also #29)

Sure. Anything that makes stuff more legible and does not break Emacs conventions too badly is a good thing in my book! :-)

Thank you again for all the work you put in here. It's excellent.

DarwinAwardWinner:
Ok, I think this is ready to merge. Make sure to read the commit message for the full details. In addition to what we've already discussed, I added proper handling of warning messages during tests. Also, in a separate commit, I made a tweak to the Makefile to run the lower-level tests in `tests/test-buttercup.el` before the higher-level ones in `docs/writing-tests.md`, since `make` stops after the first error, and the lower level tests are more helpful in debugging. In another separate commit, I added support for Markdown code blocks whose time is `emacs-lisp` or `elisp` as well as `lisp`. Markdown automatically gives such code blocks full emacs lisp syntax highlighting when editing them with `markdown-edit-code-block`.

DarwinAwardWinner:
One limitation I should mention is that currently if you have `lexical-binding` set to nil, `expect` won't work. It should be pretty easy to remove this restriction.

Another limitation is that the expression that gets printed is the fully macro-expanded version of the expression, not the one that appears literally in the call to `expect`. I think this might also be solvable.

DarwinAwardWinner:
Ok, it turns out the fix for both was about 10 lines of code total.

buttercup.el
```diff
@@ -41,6 +41,56 @@
 
 (require 'cl-lib)
 (require 'buttercup-compat)
+(require 'format-spec)
+(require 'ert nil t)
```
jorgenschaefer: Does this really need ert loaded?

buttercup.el
```diff
@@ -41,6 +41,56 @@
...
+(require 'warnings)
+
+;;;;;;;;;;;;;;;;;;;;;;;;
+;;; closure manipulation
+
+(defun buttercup--enclosed-expr (x)
+  "Given a zero-arg closure, return its unevaluated expression.
```
jorgenschaefer: Calling the argument CLOSURE would likely be more clear, no? (I'm still impressed this works even for byte-compiled stuff …)

buttercup.el
```diff
@@ -800,8 +1239,10 @@ current directory."
     (dolist (file command-line-args-left)
       (with-current-buffer (find-file-noselect file)
         (goto-char (point-min))
-        (while (re-search-forward "```lisp\n\\(\\(?:.\\|\n\\)*?\\)```"
-                                  nil t)
+        (while (let ((case-fold-search t))
```
jorgenschaefer: You can put this `let` around the `while` for minor performance wins.

buttercup.el
DarwinAwardWinner: This is mainly to get access to any ert explainers in order to give more informative messages for test failures. Note that it's a soft dependency: if ERT is unavailable for some reason, it will just fall back to the non-explainer-based failure message.

buttercup.el
DarwinAwardWinner: I'll make this change.

buttercup.el
DarwinAwardWinner: Well, the symbol `closure` already has a special meaning: it is the replacement for `lambda` in a lexical closure, although unlike `lambda`, it does not create a self-quoting expression. So it would be kind of like taking a function argument and naming it `lambda`. I think it would technically work, but it would be kind of confusing. Although now that this function works on both closures and lambdas, I guess I could just call the argument FUN. I think I'll do that.

As for working on byte-compiled things, as far as I know that shouldn't work. Figuring out how lexical scoping and macro expansion work together is tricky, but my understanding is: the `expect` macro is expanded inside a test spec. The expanded expression includes a series of lambda expressions, and is uncompiled. This happens regardless of whether buttercup.el itself is compiled. Then those lambdas are converted to closures at some point by the emacs lisp interpreter. It's possible that if you byte-compiled the test file itself, these closures would get macroexpanded and then byte-compiled and would break as a result. I'll try this and see what happens.

DarwinAwardWinner:
Ok, actually, there's some problems with this branch, but the tests are passing because of a bug in buttercup's test discovery functionality. Apparently buttercup doesn't care if a test file throws an error, as long as it defines some test specs first: https://travis-ci.org/jorgenschaefer/emacs-buttercup/jobs/264010490#L566-L567

buttercup.el
DarwinAwardWinner: Ok, this is addressed in e9eada9.

buttercup.el
DarwinAwardWinner: I've switched to more accurate naming for functions and arguments in 2f97656. The argument to this and similar functions is now FUN, and all references to closures in function names and docstrings are replaced with "function" or "wrapper function".

DarwinAwardWinner:
Ok, I think this is ready to merge now. All the tests are actually running. eeb0f6d and a19da0b address the two problems with the test suite that I found today. 02eb2fe and cad8952 are generic improvements. 2a35437 and all subsequent commits are the actual implementation.

jorgenschaefer:
Thank you very much for all this work! This is excellent :-)

PR #101, opened 2017-08-05
Allow ERT "should" and similar to work with buttercup

DarwinAwardWinner:
This defines a new macro `buttercup-with-converted-ert-signals` that converts ERT fail and skip signals to the equivalent buttercup ones. This means you should be able to use `should` and other ERT test functions inside an `it` form. This should make it easier for people to port ERT tests to buttercup.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -395,11 +395,15 @@ form.")
   "Define a spec."
   (declare (indent 1) (debug (&define sexp def-body)))
   (if body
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Very nice – and looks good to me. Do you want to squash and add a good concise commit message?

DarwinAwardWinner:
Ok, squashed.

jorgenschaefer:
Again, thank you!

PR #105, opened 2017-08-23
Let spy-on :and-return-value work with lists and symbols (close #104)

### LINKED ISSUE #104: spy-on :and-return-value evals its argument
The rewrite of spy-on in jorgenschaefer/emacs-buttercup@6ad9565cd7adc195f81bdbbc1115a6bd96802a72 broke any spy-on :and-return-value that returned a cons, list, or quoted symbol:

```emacs-lisp
(spy-on test-function :and-return-value '(a list of symbols))
```
will bind test-function to 
```emacs-lisp
(lambda () (a list of symbols))
```
which will fail with ```(invalid-function a)```

snogge:
Just a missing quote.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -574,7 +574,7 @@ responsibility to ensure ARG is a command."
             (:and-return-value
              `(lambda (&rest args)
                 ,orig-intform
-                ,arg))
+                ',arg))
             (:and-call-fake
              (let ((replacement-intform (interactive-form arg)))
                (when (and replacement-intform
--- a/tests/test-buttercup.el
+++ b/tests/test-buttercup.el
@@ -688,7 +688,47 @@
       (it "returns the specified value"
         (expect (test-function 2 3)
                 :to-equal
-                23)))
+                23))
+
+	  (it "works with strings"
+		(spy-on 'test-function :and-return-value "return value")
+		(expect (test-function 2 3)
+				:to-equal
+				"return value"))
+
+	  (it "works with vectors"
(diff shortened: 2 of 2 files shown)


vspinu:
Thanks @snogge for figuring this out. @jorgenschaefer, any chance this could be pulled soon?  We have a pile of PRs in [cider](https://github.com/clojure-emacs/cider/pulls) broken because of this.

jorgenschaefer:
Thank you very much for this fix! :-)

vspinu:
Looks like this part is still not working correctly with lambdas. Haven't dug into it, but looks like lambdas are not evaluated.

The test sample is [here](https://github.com/clojure-emacs/cider/blob/master/test/cider-apropos-tests.el#L35-L37) the output is [here](https://travis-ci.org/clojure-emacs/cider/jobs/269213398#L1680).

Any ideas?

jorgenschaefer:
There was a recent change that makes buttercup handle `(expect (cider-apropos "test") :to-throw 'user-error)` – i.e., buttercup now delays evaluation itself, without the boilerplate code. Is that the problem? If so, we should add some handling for this case for transitioning users.

vspinu:
I fixed this by replacing all lambdas with the expressions. There was actually an informative warning from buttercup all the way up of the log which I missed. So consider this one solved (at least for the cider's use case). Thanks!

jorgenschaefer:
Glad this works for you now! I'm sorry this change added more confusion ontop of the existing bug, and I hope you can enjoy buttercup without more of these annoyances now! :-D

PR #106, opened 2017-08-23
travis: Test on 25.1 and 25.2

snogge:
evm has support for Emacs 25.1 and 25.2, so why not test them too?


jorgenschaefer:
Because they weren't there when I last updated that file! Thank you very much for all your work :-)

PR #112, opened 2017-09-25
Extract function buttercup-suites-total-spces-status

snogge:
Having a single function that handles all statuses allows for
reporters using more types of status such as error for specs that
throw unexpectedly.


jorgenschaefer:
Great idea, thank you very much!

PR #113, opened 2017-10-12
Set file local variables for consistent code style

snogge:


buttercup.el
```diff
@@ -1701,3 +1701,8 @@ With buttercup minor mode active the following is activated:
 
 (provide 'buttercup)
 ;;; buttercup.el ends here
```
jorgenschaefer: This comment should mean that the file ends here, and not that it is then followed by further comments.

snogge: Updated.

buttercup.el
```diff
@@ -1701,5 +1701,10 @@ With buttercup minor mode active the following is activated:
       (cl-dolist (form imenu-forms)
         (setq imenu-generic-expression (delete form imenu-generic-expression))))))
 
+;; Local Variables:
+;; indent-tabs-mode: nil
+;; sentence-end-double-space: nil
+;; End:
 (provide 'buttercup)
 ;;; buttercup.el ends here
+
```
jorgenschaefer: At the risk of sounding highly pedantic, but there should not be an empty line at the end of the file, either :-D

snogge: I have no problem complying with that, but I am a bit surprised. Where is that policy stated?

snogge: Sorry, I see what you mean now. I'll push an update.

snogge: Updated now. Rebased the branch as well.

jorgenschaefer:
Thank you!

PR #114, opened 2017-10-12
Set failure description to empty string for pending specs

snogge:
Removes the nils from this writing-tests.md testcase

Pending specs
  can be declared using `xit'  nil
  can be declared with `it' but without a body  nil


jorgenschaefer:
Nice catch, thank you!

PR #115, opened 2017-10-13
Simplifications based on a super-struct for suite and spec

snogge:
Extracted a super struct buttercup-suite-or-spec from buttercup-suite and buttercup-spec, and then used it to simplify code and increase reuse.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -693,19 +693,11 @@ See also `buttercup-define-matcher'."
 
 (defun buttercup-suites-total-specs-defined (suite-list)
   "Return the number of specs defined in all suites in SUITE-LIST."
-  (let ((nspecs 0))
-    (dolist (spec-or-suite (buttercup--specs-and-suites suite-list))
-      (when (buttercup-spec-p spec-or-suite)
-        (setq nspecs (1+ nspecs))))
-    nspecs))
+  (length (buttercup--specs suite-list)))
 
 (defun buttercup-suites-total-specs-status (suite-list status)
   "Return the number of specs in SUITE-LIST marked with STATUS."
-  (let ((nspecs 0))
-    (dolist (spec-or-suite (buttercup--specs-and-suites suite-list) nspecs)
-      (when (and (buttercup-spec-p spec-or-suite)
-                 (eq (buttercup-spec-status spec-or-suite) status))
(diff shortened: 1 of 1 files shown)


buttercup.el
```diff
@@ -730,26 +707,32 @@ See also `buttercup-define-matcher'."
...
   (buttercup-suites-total-specs-status suite-list 'failed))
 
+(defun buttercup--specs (spec-or-suite-list)
+  "Return a flat list of all specs in SPEC-OR-SUITE-LIST."
+  (let (specs)
+    (dolist (spec-or-suite spec-or-suite-list specs)
+      (if (buttercup-spec-p spec-or-suite)
+          (setq specs (append specs (list spec-or-suite)))
+        (setq specs (append specs (buttercup--specs
+                                   (buttercup-suite-children spec-or-suite))))))))
```
jorgenschaefer: Not that it matters much performance-wise, but this is inefficient and highly unidiomatic – could you use the usual push/nreverse or (better, I think) the cl-loop pattern? :-)

buttercup.el
```diff
@@ -730,26 +707,32 @@ See also `buttercup-define-matcher'."
...
 (defun buttercup-suite-full-name (suite)
   "Return the full name of SUITE, which includes the names of the parents."
-  (let ((name-parts (mapcar #'buttercup-suite-description
-                            (cons suite (buttercup-suite-parents suite)))))
-    (mapconcat #'identity
-               (reverse name-parts)
-               " ")))
+  (mapconcat #'buttercup-suite-description
+             (nreverse (cons suite (buttercup-suite-parents suite)))
+             " "))
```
jorgenschaefer: Woah, nice catch!

buttercup.el
snogge: I think I tried both approaches and the problem is keeping the order.  buttercup--specs-and-suites is equally unappealing.

I'll have a think and see if I can come up with something.

jorgenschaefer: Hm. Ah yes, tree recursive stuff. I think I'm actually fine with this then. Thank you!

jorgenschaefer:
I'm happy to merge this now – anything you want to add or change in the branch still?

snogge:
I have nothing more.

jorgenschaefer:
Thank you!

PR #117, opened 2017-10-21
Change shebang to use /usr/bin/env

etu:
Not all systems provides a /bin/bash (even GNU/Linux based systems!), but /usr/bin/env is present on pretty much all systems that I can think of.

For example: I'm on NixOS. NixOS places everything that matters in other places than `/bin` and `/usr/bin`. The only path within `/bin` on my system is `/bin/sh` which is a symlink. And the only path within `/usr` on my system is `/usr/bin/env` which (surprise surprise) is also a symlink to `/nix/...somewhere...`

### DIFF
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ -n "$INSIDE_EMACS" ]
 then


jorgenschaefer:
I don't see a problem to include this PR, but it's going to be an uphill battle for NixOS to be different than everything else :-D

```
$ grep '#!/bin/bash' /usr/bin/* | wc -l
37
```

etu:
@jorgenschaefer Thanks!

Well, it's not as bad as you think since NixOS has a function that can automaticly fix up shebangs during the fixup-phase of the packaging process. But, this package is installed by Cask, not by Nix. Hence nix can't fix it that easily. It's probably possible to package buttercup for nix, but that seemed like more work ;)

Issue #121, opened 2018-02-11
Is Windows 10 supported?


chrycheng:
I could not find documentation but there seems to be some Windows support since there's a BAT file.  However, I get the following error when I run `cask exec buttercup -L .`:

```
c:\my-proj>cask exec buttercup -L .

c:\my-proj>/bin/bash: c:my-proj.cask25.2elpabuttercup-1.9binbuttercup: No such file or directory
```

It looked to me like the path needed to be translated into Unix so I tried to fix it by changing `c:\my-proj\.cask\25.2\elpa\buttercup-1.9\bin\buttercup.bat` to:

```bat
@echo off
set original=%~dp0buttercup
set fixedSlash=%original:\=/%
set fixedDrive=/mnt/%fixedSlash::=%
@bash %fixedDrive%
```

This changed the error to:

```
c:\my-proj>cask exec buttercup -L .

c:\my-proj>Cannot open load file: no such file or directory, buttercup
```

It would seem that the `EMACSLOADPATH` (and `PATH`) environment variables weren't passed to the Bash process.

jorgenschaefer:
I merged your fix, thank you :-)

PR #123, opened 2018-02-12
Fix issue #121

chrycheng:


jorgenschaefer:
Thank you so much for your work! This is excellent :-)

I do not have access to a Windows machine, so I can't test buttercup on windows. Having someone do that is excellent!

PR #128, opened 2018-07-23
WIP: Add support for capturing dynamic binding environment

Fuco1:
This is quite a bit of black magic and I'm using eval, however I can't really find any other way.

Without using `:env` it falls back to the old code so this should not affect any current usage.


buttercup.el
```diff
@@ -675,6 +675,7 @@ See also `buttercup-define-matcher'."
 (cl-defstruct (buttercup-suite (:include buttercup-suite-or-spec))
   ;; Any children of this suite, both suites and specs
   children
+  environment
```
Fuco1: This is probably unnecessary, we can track the environment in the `buttercup--current-env` variable (seems like a duplicate)

buttercup.el
```diff
@@ -783,27 +784,38 @@ See also `buttercup-define-matcher'."
...
+         (env (plist-get body :env))
+         (body (progn
+                 (while (memq (car body) (list :var :env))
+                   (setq body (cddr body)))
+                 body))
+         (new-body (if var
+                       `((let ,var
+                           ,@body))
+                     body)))
+    `(buttercup-describe ,description (lambda () ,@new-body) ',env)))
```
Fuco1: Passing `',env` here is really ugly... any ideas?

buttercup.el
```diff
@@ -851,7 +863,31 @@ function containing the body instructions passed to `it'."
...
+                                                 (if (consp binding)
+                                                     (cons (car binding) (symbol-value (car binding)))
+                                                   (cons binding (symbol-value binding))))
+                                               (buttercup-suite-environment suite))))
+                                         (unwind-protect
+                                             (progn
+                                               (mapc
+                                                (lambda (binding)
+                                                  (if (consp binding)
+                                                      (setf (symbol-value (car binding)) (eval (cadr binding)))
```
Fuco1: `eval` :(  But, since these are functions and we are passing data I don't see any good way, other than walking the trees in `describe` macro and looking for `it` forms and installing let-bindings there.

Fuco1:
What you described is a common idiom in javascript but remember that there the scoping works different

```js
describe('getProductValue', () => {

    let edsnDown

    describe('when edsn is up', () => {

        beforeEach(() => {
            edsnDown = false
        })

        it("should return '1' when we have elk only", () => {
            expect(getProductValue(true, edsnDown)).toEqual('1')
        })

        it("should return '2' when we have elk and gas", () => {
            expect(getProductValue(false, edsnDown)).toEqual('2')
        })

    })

})
```

here the `edsnDown` will never spill outside of the describe... in Emacs special variables will :(

jorgenschaefer:
How about `around`?

```Lisp
(describe "Backup"
  (around-each block
    (let ((kill-ring '("foobar")))
      (funcall block)))

  (it "kill ring should not be empty"
    (expect (< 0 (length kill-ring)) :to-be-truthy)))
```

Fuco1:
@jorgenschaefer That is actually really cool idea.  You were doing some ruby lately heh? :D

Yea I think that could work if implemented properly.  I'll try to take a look on that if you don't mind (or have it already implemented)

jorgenschaefer:
Am I getting a Ruby dialect? Help! :-D I have nothing implemented and I would love such a feature, thank you!

bbatsov:
@Fuco1 Great work so far with this! Looking forward to its completion! 

P.S. As a Rubyist myself I welcome any suggestion to get buttercup closer to what RSpec offers - it's a pretty amazing framework.

Fuco1:
I probably won't work on this in any reasonably close future.

snogge:
This PR is probably older than my maintainership, and as it was marked WIP I never gave it much thought.
I don't think I'll get to it soon considering how little time I have to spend on this project.  
If you have any specific problem, please file a separate issue. Maybe we can figure something out.

PR #129, opened 2018-08-22
Fix obsolete aliases

snogge:
We could change the obsolete aliases to normal aliases, it could make some things easier or prettier.


jorgenschaefer:
Yes indeed! Thank you :-)

PR #133, opened 2018-08-28
Keep specs in pending suites, fixes #116 

### LINKED ISSUE #116: specs in pending suites should be processed
Any suites and specs nested in a xdescribe form will be skipped during processing and never added to buttercup-suites.  This means they are not counted and not matchable with patterns.

For instance, the command

```ShellSession
emacs -batch -L . -l buttercup.el -f buttercup-run-markdown docs/writing-tests.md
```

in the emacs-buttercup folder will print 

    Ran 50 out of 52 specs, 0 failed, in 0.0 seconds.

while there is actually a 53rd spec at line 318 of writing-tests.md that is skipped and not counted.

If a matching pattern should match any spec in a xdescribe form, that spec will not be found and the containing, skipped, xdescribe will not be run either.

snogge:
This is my attempt at fixing #116 without introducing new complexity.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -965,18 +965,6 @@ mainly calls to `describe', `it' and `before-each'."
      ;; make sure the suite is marked as pending
      (signal 'buttercup-pending "PENDING")))
 
-(defun buttercup-xdescribe (description function)
-  "Like `buttercup-describe', but mark the suite as disabled.
-
-A disabled suite is not run.
-
-DESCRIPTION has the same meaning as in `xdescribe'. FUNCTION
-is ignored.
-`describe'."
-  (ignore function)
-  (buttercup-describe description (lambda ()
-                                    (signal 'buttercup-pending "PENDING"))))
-
 ;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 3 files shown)


buttercup.el
```diff
@@ -915,27 +915,55 @@ FUNCTION is a function containing the body instructions passed to
...
-  "Like `describe', but mark the suite as disabled.
+(defun buttercup--disable-specs (forms)
+  "Process FORMS to make any suites or specs pending."
+  (when (eq (car forms) :var)
+    (setq forms (cddr forms)))
+  (let (retained inner)
+    (dolist (form forms (nreverse retained))
+      (pcase form
+        ;; Make it pending by just keeping the description
+        (`(it ,description . ,_)
```
snogge: These two cases can be joined into 
```
(`(,(or 'it 'xit) ,description . ,_)
   (push (list 'it description) retained))
```
if it wasn't for Emacs 24.

jorgenschaefer:
You could probably just ignore everything excelt `x?it` and `x?describe`, withough prescribing what that is – what kind of "unrecognized form" do you expect?

PR #134, opened 2018-09-02
Set failure-description for pending specs

snogge:
The status slot in buttercup-suite-or-spec is set ahead of time for
pending (not skipped) specs.  This meant that the failure-description
and failure-stack of pending specs were never set in
buttercup--update-with-funcall.  This in turn meant that the output of
pending specs was

    a pending spec

rather than

    a pending spec  PENDING

which it should have been.

The failure-stack will still always be nil for pending specs.


jorgenschaefer:
Thank you very much! :-)

PR #135, opened 2018-09-10
Timed specs

snogge:
The "Print elapsed time for each spec" commit is a demonstration of what can be done with the timestamps added in the commit before. The elapsed time output may need some more formatting.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -670,7 +670,8 @@ See also `buttercup-define-matcher'."
   (status 'passed)
   failure-description
   failure-stack
-  )
+  time-started
+  time-ended)
 
 (cl-defstruct (buttercup-suite (:include buttercup-suite-or-spec))
   ;; Any children of this suite, both suites and specs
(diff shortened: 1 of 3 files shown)


buttercup.el
```diff
@@ -670,7 +670,9 @@ See also `buttercup-define-matcher'."
   (status 'passed)
   failure-description
   failure-stack
-  )
+  ;; timestamps
```
jorgenschaefer: This comment is not needed :-)

snogge: Oops, removed.

snogge:
I'm ready to merge this if you have no other comments.

snogge:
By the way, it would be great if you could bump the version after merging this.

jorgenschaefer:
Thank you for your contribution! There are regular releases (about once a month), I hope that's ok :-)

snogge:
I was not aware of the releases schedule.  I can wait, no problem.

Fuco1:
My gods thank you!  I'm just in the process of migrating smartparens to buttercup and I was wondering how I can get this behaviour (ert-runner has it)... I installed buttercup and it was just there :D <3

PR #144, opened 2019-08-27
Add support for :var*

### LINKED ISSUE #141: :var should work like let*
This doesn't work:
```el
(describe "org-ql"
  :var ((test-buffer (find-file-noselect (concat default-directory "tests/data.org")))
        (num-headings (with-current-buffer test-buffer
                        (org-with-wide-buffer
                         (goto-char (point-min))
                         (cl-loop while (re-search-forward org-heading-regexp nil t)
                                  sum 1)))))
```
I get this error:
```
Symbol’s value as variable is void: test-buffer
```
This doesn't work either:
```el
(describe "org-ql"
  :var* ((test-buffer (find-file-noselect (concat default-directory "tests/data.org")))
         (num-headings (with-current-buffer test-buffer
                         (org-with-wide-buffer
                          (goto-char (point-min))
                          (cl-loop while (re-search-forward org-heading-regexp nil t)
                                   sum 1)))))
```
I get this error:
```
Invalid function: (test-buffer (find-file-noselect (concat default-directory "tests/data.org")))
```
I could use `setq` in a `before-all` block instead of `:var`, but that (sometimes) causes `Unused lexical variable` warnings all over the Buttercup output.  I haven't figured out exactly why or when.

Am I doing something wrong?

Thanks.

kevinjfoley:
Wraps suite in let* form

I believe this should fix #141

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
   - EVM_EMACS=emacs-25.1-travis
   - EVM_EMACS=emacs-25.2-travis
   - EVM_EMACS=emacs-26.1-travis
+  - EVM_EMACS=emacs-26.2-travis
 before_install:
   - curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > travis.sh && source ./travis.sh
   - evm install "$EVM_EMACS" --use --skip
--- a/buttercup.el
+++ b/buttercup.el
@@ -805,10 +805,15 @@ form.")
 DESCRIPTION is a string. BODY is a sequence of instructions,
 mainly calls to `describe', `it' and `before-each'."
   (declare (indent 1) (debug (&define sexp def-body)))
-  (let ((new-body (if (eq (elt body 0) :var)
-                      `((let ,(elt body 1)
-                          ,@(cddr body)))
-                    body)))
+  (let ((new-body
+         (cond
+          ((eq (elt body 0) :var)
+           `((let ,(elt body 1)
+               ,@(cddr body))))
+          ((eq (elt body 0) :var*)
+           `((let* ,(elt body 1)
+               ,@(cddr body))))
+          (t body))))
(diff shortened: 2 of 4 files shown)


snogge:
Could you add some tests and update the docs too, please?

snogge:
What version of Emacs are you using when the test in 2 fails?

snogge:
So the miscount (1) is because of the "The `buttercup-run' function" description, there is still something there that replaces the main list of suites and specs with what is supposed to be local test data.  The bug is in that testcase, not in emacs-buttercup. 

For (2), the test passes for me with 
GNU Emacs 26.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.11)
 of 2019-08-27.
But looking at the test now (and I must admit to writing it), it is **very** hard for me to break.  The test is just to lenient.  What is the error message?

snogge:
> " to match the regexp "([0-9]+\\.[0-9]+\\(h\\|m\\|m?s\\))", but instead it was "  nil (0s)

Ha, your machine is too fast :) 
That should be easy to fix.

Anyway, neither of the problems you have reported is a result of your pull request, but all my fault :( 
I will post fixes for them tonight.  Over to you @jorgenschaefer .

snogge:
As you can see the fixes for the tests are in #145, it would be great if you could try them out.

jorgenschaefer:
Hello! Thank you for all your work :-) I am sadly woefully underequipped with time to keep up with the discussions here – I added @snogge as a collaborator, please do merge your PRs once you are happy with them, ok? Thanks again for your work!

snogge:
Thank you @jorgenschaefer.  Will you still do the releases when necessary?

jorgenschaefer:
I just did a release, but you should be able to do the same by just running `./scripts/release` – these days, it's as simple as adding a tag, luckily! :-)

PR #146, opened 2019-08-29
Make spies record calls that throw errors

### LINKED ISSUE #139: Called times doesn't increase if error was thrown
```elisp
(defun foo () "anything")
(describe "Called times increases"
          (it "when error was thrown"
              (spy-on 'foo :and-throw-error 'error)
              (expect (foo) :to-throw 'error)
              (expect 'foo :to-have-been-called-times 1)))
```
This test fails for me with buttercup 1.16.

DarwinAwardWinner:
Fixes #139.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1195,6 +1195,16 @@ responsibility to ensure ARG is a command."
   "Return the number of times SPY has been called so far."
   (length (spy-calls-all spy)))
 
+(defun spy-calls-count-returned (spy)
+  "Return the number of times SPY has been called successfully so far."
+  (length (cl-remove-if-not 'spy-context-return-p
+                            (spy-calls-all spy))))
+
+(defun spy-calls-count-errors (spy)
+  "Return the number of times SPY has been called and thrown errors so far."
+  (length (cl-remove-if-not 'spy-context-thrown-p
+                            (spy-calls-all spy))))
+
 (defun spy-calls-args-for (spy index)
   "Return the context of the INDEXth call to SPY."
   (let ((context (elt (spy-calls-all spy)
--- a/buttercup.el
+++ b/buttercup.el
@@ -647,7 +647,7 @@ See also `buttercup-define-matcher'."
   (setq spy (funcall spy)
         number (funcall number))
   (cl-assert (symbolp spy))
-  (let* ((call-count (length (spy-calls-all spy))))
+  (let* ((call-count (spy-calls-count spy)))
     (cond
      ((= number call-count)
       t)
--- a/buttercup.el
+++ b/buttercup.el
@@ -1094,32 +1094,32 @@ responsibility to ensure ARG is a command."
                 nil))
             (_
              (error "Invalid `spy-on' keyword: `%S'" keyword)))))
-    (buttercup--spy-on-and-call-fake symbol replacement)))
+    (buttercup--spy-on-and-call-replacement symbol replacement)))
 
-(defun buttercup--spy-on-and-call-fake (spy fake-function)
-  "Replace the function in symbol SPY with a spy calling FAKE-FUNCTION."
+(defun buttercup--spy-on-and-call-replacement (spy fun)
+  "Replace the function in symbol SPY with a spy calling FUN."
   (let ((orig-function (symbol-function spy)))
-    (fset spy (buttercup--make-spy fake-function))
+    (fset spy (buttercup--make-spy fun))
     (buttercup--add-cleanup (lambda ()
                               (fset spy orig-function)))))
 
-(defun buttercup--make-spy (fake-function)
-  "Create a new spy function wrapping FAKE-FUNCTION and tracking calls to itself."
+(defun buttercup--make-spy (fun)
+  "Create a new spy function wrapping FUN and tracking calls to itself."
   (let (this-spy-function)
     (setq this-spy-function
           (lambda (&rest args)
-            (let ((return-value (apply fake-function args)))
+            (let ((return-value (apply fun args)))
               (buttercup--spy-calls-add
                this-spy-function
                (make-spy-context :args args
                                  :return-value return-value
                                  :current-buffer (current-buffer)))
               return-value)))
-    ;; Add the interactive form from `fake-function', if any
-    (when (interactive-form fake-function)
+    ;; Add the interactive form from `fun', if any
+    (when (interactive-form fun)
       (setq this-spy-function
             `(lambda (&rest args)
-               ,(interactive-form fake-function)
+               ,(interactive-form fun)
                (apply ',this-spy-function args))))
     this-spy-function))
 
--- a/buttercup.el
+++ b/buttercup.el
@@ -1133,8 +1133,8 @@ responsibility to ensure ARG is a command."
                                  :return-value nil
                                  :thrown-signal err
                                  :current-buffer (current-buffer))))
-            ;; Regardless, we only caught this error order to record
-            ;; it, so we need to re-throw it.
+            ;; Regardless, we only caught this error in order to
+            ;; record it, so we need to re-throw it.
             (signal (car err) (cdr err)))))))
     ;; Add the interactive form from `fun', if any
     (when (interactive-form fun)
--- a/buttercup.el
+++ b/buttercup.el
@@ -1025,6 +1025,7 @@ DESCRIPTION has the same meaning as in `xit'. FUNCTION is ignored."
 (cl-defstruct spy-context
   args
   return-value
+  thrown-signal
   current-buffer)
 
 (defun spy-on (symbol &optional keyword arg)
@@ -1106,15 +1107,35 @@ responsibility to ensure ARG is a command."
 (defun buttercup--make-spy (fun)
   "Create a new spy function wrapping FUN and tracking calls to itself."
   (let (this-spy-function)
-    (setq this-spy-function
-          (lambda (&rest args)
-            (let ((return-value (apply fun args)))
+    (setq
+     this-spy-function
+     (lambda (&rest args)
+       (let ((returned nil)
+             (return-value nil))
+         (condition-case err
+             (progn
+               (setq return-value (apply fun args)
+                     returned t)
+               (buttercup--spy-calls-add
+                this-spy-function
+                (make-spy-context :args args
+                                  :return-value return-value
+                                  :thrown-signal nil
+                                  :current-buffer (current-buffer)))
+               return-value)
+           (error
+            ;; If returned is non-nil, then the error we caught
+            ;; didn't come from FUN, so we shouldn't record it.
+            (unless returned
               (buttercup--spy-calls-add
                this-spy-function
                (make-spy-context :args args
-                                 :return-value return-value
-                                 :current-buffer (current-buffer)))
-              return-value)))
+                                 :return-value nil
+                                 :thrown-signal err
+                                 :current-buffer (current-buffer))))
+            ;; Regardless, we only caught this error order to record
+            ;; it, so we need to re-throw it.
+            (signal (car err) (cdr err)))))))
     ;; Add the interactive form from `fun', if any
     (when (interactive-form fun)
       (setq this-spy-function
--- a/buttercup.el
+++ b/buttercup.el
@@ -1023,10 +1023,14 @@ DESCRIPTION has the same meaning as in `xit'. FUNCTION is ignored."
   "A mapping of currently-defined spies to their contexts.")
 
 (cl-defstruct spy-context
-  args
-  return-value
-  thrown-signal
-  current-buffer)
+  args current-buffer)
+;; The struct and slot names are kind of a cheat so that the accessor
+;; function names remain unchanged: `spy-context-return-value' and
+;; `spy-context-thrown-signal'.
+(cl-defstruct (spy-context-return (:include spy-context))
+  value)
+(cl-defstruct (spy-context-thrown (:include spy-context))
+  signal)
 
 (defun spy-on (symbol &optional keyword arg)
   "Create a spy (mock) for the function SYMBOL.
@@ -1118,10 +1122,9 @@ responsibility to ensure ARG is a command."
                      returned t)
                (buttercup--spy-calls-add
                 this-spy-function
-                (make-spy-context :args args
-                                  :return-value return-value
-                                  :thrown-signal nil
-                                  :current-buffer (current-buffer)))
+                (make-spy-context-return :args args
+                                         :value return-value
+                                         :current-buffer (current-buffer)))
                return-value)
            (error
             ;; If returned is non-nil, then the error we caught
@@ -1129,10 +1132,9 @@ responsibility to ensure ARG is a command."
             (unless returned
               (buttercup--spy-calls-add
                this-spy-function
-               (make-spy-context :args args
-                                 :return-value nil
-                                 :thrown-signal err
-                                 :current-buffer (current-buffer))))
+               (make-spy-context-thrown :args args
+                                        :signal err
+                                        :current-buffer (current-buffer))))
             ;; Regardless, we only caught this error in order to
             ;; record it, so we need to re-throw it.
             (signal (car err) (cdr err)))))))
--- a/tests/test-buttercup.el
+++ b/tests/test-buttercup.el
@@ -619,8 +619,6 @@
     ;; We use `before-all' here because some tests need to access the
     ;; same function as previous tests in order to work, so overriding
     ;; the function before each test would invalidate those tests.
-    ;; Unfortunately there's no way to do this lexically, so these
-    ;; function definitions are leaked after the tests run.
     (before-all
       (setq saved-test-function (and (fboundp 'test-function)
                                      (symbol-function 'test-function))
@@ -644,8 +642,8 @@
           (fset 'test-command saved-test-command)
         (fmakunbound 'test-command))
       (if saved-test-function-throws-on-negative
-          (fset 'test-test-function-throws-on-negative
-                test-function-throws-on-negative)
+          (fset 'test-function-throws-on-negative
+                saved-test-function-throws-on-negative)
         (fmakunbound 'test-function-throws-on-negative)))
 
     (describe "`spy-on' function"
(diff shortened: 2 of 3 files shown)


DarwinAwardWinner:
Technically this is changing the semantics of the spy functionality in a backward-incompatible way. It's conceivable that people would be relying on e.g. the count of calls actually being the count of *ssuccessful* calls. In practice, I doubt anyone did this, so this probably won't break anyone's tests.

buttercup.el
```diff
@@ -1094,32 +1094,32 @@ responsibility to ensure ARG is a command."
                 nil))
             (_
              (error "Invalid `spy-on' keyword: `%S'" keyword)))))
-    (buttercup--spy-on-and-call-fake symbol replacement)))
+    (buttercup--spy-on-and-call-replacement symbol replacement)))
 
-(defun buttercup--spy-on-and-call-fake (spy fake-function)
-  "Replace the function in symbol SPY with a spy calling FAKE-FUNCTION."
+(defun buttercup--spy-on-and-call-replacement (spy fun)
```
snogge: Wouldn't `(defun buttercup--spy-on (fun mock)` be more appropriate?

tests/test-buttercup.el
```diff
@@ -613,264 +613,265 @@
...
-                 :to-have-been-called-times
-                 (make-list-of-closures '(test-function 3)))
-                :to-equal
-                (cons nil
-                      "Expected `test-function' to have been called 3 times, but it was called 2 times")))
+  ;; We use `before-all' here because some tests need to access the
+  ;; same function as previous tests in order to work, so overriding
+  ;; the function before each test would invalidate those tests.
+  ;; Unfortunately there's no way to do this lexically, so these
+  ;; function definitions are leaked after the tests run.
```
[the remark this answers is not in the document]
DarwinAwardWinner: Is it possible to lexically bind a symbol's function slot? I don't actually know.

tests/test-buttercup.el
```diff
@@ -619,11 +619,15 @@
   ;; Unfortunately there's no way to do this lexically, so these
   ;; function definitions are leaked after the tests run.
   (before-all
-    (fset 'test-function (lambda (a b)
-                           (+ a b)))
-    (fset 'test-command (lambda ()
-                          (interactive)
-                          t)))
+    (fset 'test-function
```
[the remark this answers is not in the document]
DarwinAwardWinner: I think I did it to be consistent with the new test function I added below, whose name was too long to put the definition on the same line. But I'll revert it.

tests/test-buttercup.el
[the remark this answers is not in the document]
snogge: I tested this in *scratch*: 
```
(cl-letf (((symbol-function 'my-test-function) nil))
  (fset 'my-test-function (lambda () "1"))
  (my-test-function))
"1"
(my-test-function) => (void-function my-test-function)
```
the macroexpansion is actually rather straightforward.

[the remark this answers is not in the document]
DarwinAwardWinner: Ok, I'll add that in, then.

buttercup.el
DarwinAwardWinner: The first arg is a symbol, not a function, and other functions consistently use `spy` for the argument name of the symbol whose function is being spied on, so I don't think it makes sense to change it here. As for the second arg name and the function name, I'm happy to call them whatever you think makes the most sense.

tests/test-buttercup.el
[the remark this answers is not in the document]
snogge: Quick test in real code shows it wasn't that simple :(
It's probably as you say, it's not possible to lexically bind the function slot.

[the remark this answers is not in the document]
DarwinAwardWinner: Actually, looking at the macro expansion, isn't that dynamic binding? It sets the symbol function at the start of the block and then reverts it to its old value when the block exits. Since the `describe` block only defines the test and doesn't run it, the dynamic binding will be long gone when the tests actually run, having no effect.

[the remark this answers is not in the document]
DarwinAwardWinner: Wasn't someone working on an `around-all` macro or something? I think that would probably be what's needed here.

buttercup.el
snogge: Yeah, I noticed that it is called spy elsewhere and I find the naming confusing all around. But lets not bikeshed it, keep you change as it is.

tests/test-buttercup.el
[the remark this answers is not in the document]
snogge: I was hoping to not have to reindent that big chunk of code :)
But never mind, we can go ahead with this part of the change as well.

snogge:
Have you checked if the docs need updateing?

DarwinAwardWinner:
Yeah, I guess this is a new feature that ought to be documented. I was waiting to see if you were interested in merging it before I put in the effort to write the docs.

snogge:
I think we should merge it. I guess it's the "Other tracking properties" section that needs some work.

DarwinAwardWinner:
Ok, so really I should write docs for examining the return value as well, presumably using `:and-call-through` since that's the only time inspecting the return value or error code will be informative.

tests/test-buttercup.el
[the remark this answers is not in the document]
snogge: I pushed an alternative to #147, what do you think?  It's not pretty, but it does not "leak" and test functions.
Should be possible to convert to some sort of `:func`argument to `describe`.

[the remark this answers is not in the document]
DarwinAwardWinner: Yes, I think that should work. Perhaps instead of `:func` you could have a `:letf` argument to `describe` that wraps the test-running code in the equivalent `cl-letf` form? Alternatively, I think the most general solution would be to implement `around-each` and `around-all` macros. I might take a stab at that some time soon. For now, though, I think your solution is enough.

DarwinAwardWinner:
Actually, the entire spy-context struct seems undocumented, other than the fact that some functions return a context, and you can create one using `make-spy-context`. Maybe that should be a separate issue.

tests/test-buttercup.el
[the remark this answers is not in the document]
DarwinAwardWinner: I just realized `cl-flet` and `cl-labels` are supposed to be lexically scoped. (They actually work by rewriting the function names to different, unused, symbols.) I'll test things out using those and see if it works.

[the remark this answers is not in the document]
DarwinAwardWinner: Never mind, these actually work by lexically binding symbols normally using `let` and then rewriting calls to the bound functions to use `funcall` to access the symbol value rather than the symbol function. So that won't work.

DarwinAwardWinner:
I've rebased this onto the latest master and added some additional tests and exposition in the docs. It should be ready to merge.

buttercup.el
```diff
@@ -1106,15 +1107,35 @@ responsibility to ensure ARG is a command."
...
+            ;; If returned is non-nil, then the error we caught
+            ;; didn't come from FUN, so we shouldn't record it.
+            (unless returned
               (buttercup--spy-calls-add
                this-spy-function
                (make-spy-context :args args
-                                 :return-value return-value
-                                 :current-buffer (current-buffer)))
-              return-value)))
+                                 :return-value nil
```
snogge: `nil` could be a valid return value. Tests could misstake that for valid returns unless the test is careful to check for thrown signals.
I suppose we could use a symbol like `buttercup-noreturn' or something.  It seems unlikely that a function outside buttercup would return that value in normal circumstances.

buttercup.el
```diff
@@ -1106,15 +1107,35 @@ responsibility to ensure ARG is a command."
...
+                                 :thrown-signal err
+                                 :current-buffer (current-buffer))))
+            ;; Regardless, we only caught this error order to record
```
snogge: I think there is a missing 'in' between 'error' and 'order'.

tests/test-buttercup.el
```diff
@@ -613,27 +613,40 @@
...
 (describe "The Spy "
-  (let (saved-test-function saved-test-command)
+  (let (saved-test-function
+        saved-test-command
+        saved-test-function-throws-on-negative)
     ;; We use `before-all' here because some tests need to access the
     ;; same function as previous tests in order to work, so overriding
     ;; the function before each test would invalidate those tests.
+    ;; Unfortunately there's no way to do this lexically, so these
+    ;; function definitions are leaked after the tests run.
```
snogge: Do they? Isn't `fmakunbound` in the after-each enough to remove them?

tests/test-buttercup.el
```diff
@@ -613,27 +613,40 @@
...
     (after-all
       (if saved-test-function
           (fset 'test-function saved-test-function)
         (fmakunbound 'test-function))
       (if saved-test-command
           (fset 'test-command saved-test-command)
-        (fmakunbound 'test-command)))
+        (fmakunbound 'test-command))
+      (if saved-test-function-throws-on-negative
+          (fset 'test-test-function-throws-on-negative
```
snogge: One `test-` too many here.

tests/test-buttercup.el
```diff
@@ -613,27 +613,40 @@
...
+                test-function-throws-on-negative)
```
snogge: And shouldn't this be `saved-test-function-throws-on-negative`?

tests/test-buttercup.el
```diff
@@ -881,7 +894,52 @@
...
-                'error "Stubbed error")))))
+                'error "Stubbed error")))
+
+    (describe "error-recording functionality"
+      (before-each
+        (spy-on 'test-function-throws-on-negative :and-call-through))
+
+      (it "records the function as called even if it throws an error"
+        (expect (test-function-throws-on-negative -5) :to-throw)
+        (expect (buttercup--apply-matcher
```
snogge: Why are you not using `(expect 'test-function-throws-on-negative :to-have-been-called)` here?
Same for the cases below.

tests/test-buttercup.el
```diff
@@ -881,7 +894,52 @@
...
+                :to-be
+                t))
+
+      (it "records the signal thrown by a call to the function"
+        (test-function-throws-on-negative 5)
+        (expect (test-function-throws-on-negative -5) :to-throw 'error)
+        (expect (spy-context-thrown-signal
+                 (spy-calls-first 'test-function-throws-on-negative))
+                :to-be nil)
+        (expect (car
```
snogge: Why not check the entire signal like you do in the example in the next commit.

buttercup.el
DarwinAwardWinner: There's always the option to use `gensym` to make an unique symbol that's guaranteed not to be in use anywhere else.

tests/test-buttercup.el
DarwinAwardWinner: Yes, that comment is left over from an earlier version of the tests that did leak the function definitions. I'll delete it

buttercup.el
snogge: That sounds like a better idea.  You should only use the `thrown-signal` slot to determine whether a call has thrown an error.

tests/test-buttercup.el
DarwinAwardWinner: If you look above, many of the existing test cases do likewise. I don't recall the reason at the moment, but I vaguely remember that there *is* a reason. In any case, I'm just following the pattern.

buttercup.el
DarwinAwardWinner: Returning any arbitrary symbol isn't great either. If the caller just checks whether the return value is non-nil, the test will pass. With the current design, the onus is entirely on the caller to check the thrown signal in order to determine whether the function failed.

Would it be reasonable for `spy-context-return-value` to throw an error when called on a spy context for a function call that threw an error? (And is this implementable on top of the `cl-defstruct`?)

tests/test-buttercup.el
DarwinAwardWinner: Fixed in https://github.com/jorgenschaefer/emacs-buttercup/pull/146/commits/046031076bd7a53ab4d5a6dd873941a18c7d612f.

tests/test-buttercup.el
DarwinAwardWinner: Fixed in 0460310.

tests/test-buttercup.el
DarwinAwardWinner: Fixed in 0460310.

buttercup.el
DarwinAwardWinner: Fixed in 0460310.

buttercup.el
snogge: Do you think it would work to use different structs for calls that return and calls that throw?
```emacs-lisp
(cl-defstruct spy-context args current-buffer)
(cl-defstruct (spy-context-return :include spy-context) return-value)
(cl-defstruct (spy-context-error :include spy-context) thrown-signal)
```
Then it would not be possible to get the `return-value` of a call that threw an error. No idea if that has weird consequences somewhere.

As an aside you might want to add functions `spy-calls-count-returned` and  `spy-calls-count-errors` or similar.

DarwinAwardWinner: I think that is, practically speaking, equivalent to my proposal to have `spy-context-return-value` throw an error if the call threw an error, since that's exactly what will happen if you call it on a spy-context-error struct. However, we should probably not implement it in this way, because we probably want to provide our own error message.

Of course, my proposal means we need to somehow redefine `spy-context-return-value` after the struct is defined. I guess we can do that with advice. Having a package define advice for its own function is kind of odd, but it sort of makes sense here, given that I don't think there's any other way to reliably redefine a struct accessor function.

DarwinAwardWinner: Regardless, I think I'm even further convinced that the only reasonable thing to do is to throw an error if a test tries to access the return value of a failed call. This will cause the test to fail, which is probably what we want if a function call is failing in a way that the test didn't expect.

DarwinAwardWinner: Any opinions on whether advising a function that was defined by `cl-defstruct` is an appropriate way to implement this? @snogge

snogge: I'm not big on using advice this way, but I guess it is the least complicated way forward.

I still think the code below could be tweaked to do more or less the same and give room for future expansion.  But maybe it's too complicated?
```emacs-lisp
(cl-defstruct spy-context args current-buffer)
(cl-defstruct (spy-context--return (:include spy-context)) value)
(cl-defstruct (spy-context--thrown (:include spy-context)) signal)
;; Not sure when cl-defgeneric was introduced
(defgeneric spy-context-return-value (context)
  (error "Context is not a return"))
(defmethod spy-context-return-value ((context spy-context--return))
  (spy-context--return-value context))
(defgeneric spy-context-thrown-signal (context)
  (error "Context is not an error"))
(defmethod spy-context-thrown-signal ((context spy-context--thrown))
  (spy-context--thrown-signal context))
```

DarwinAwardWinner: Yeah, that looks good. I'll try to find out when `cl-defgeneric` was added to see if we can use it or not.

DarwinAwardWinner: Looks like `cl-defgeneric` was added in 25.1. I guess that means we can't use it if we're supporting all the way back to 24.3.

DarwinAwardWinner: Ok, this is now implemented using your suggestions, including `spy-calls-count-returned` and `spy-calls-count-errors`.

DarwinAwardWinner:
Test failures on 26.x seem spurious:

>     $ make test
>     emacs -batch -L . -f batch-byte-compile buttercup-compat.el
>     /tmp/emacs-26.1-travis/bin/emacs: error while loading shared libraries: libgnutls.so.26: cannot open shared object file: No such file or directory
>     Makefile:17: recipe for target 'buttercup-compat.elc' failed
>     make: *** [buttercup-compat.elc] Error 127

The command "make test" exited with 2.

snogge:
The problem with libgnutls has been fixed on master if you want to rebase or merge from there.

I'm OK with merging this now. I tested to merge and build with travis, and there are no merge conflicts and all tests pass.

Is there anything you still want to do with this PR?

DarwinAwardWinner:
I just noticed the text in writing-tests.md still describes the old struct instead of the different structs for returns and errors. I'll quickly update it, and rebase onto latst master.

DarwinAwardWinner:
Ok, should be good to go now. However, one thing to pay attention to is backward compatibility. I *think* this should actually be backward-compatible thanks to some clever struct naming shenanigans. The old struct was called `spy-context` and had a slot named `return-value`, so the compiler-generated accessor was named `spy-context-return-value`. The new struct is called `spy-context-return` with a slot named `value`, which generates the same accessor name. So any existing code calling `spy-context-return-value` should continue to work. Meanwhile, the 2 common slots are still accessed via the accessors defined in the `spy-context` struct (which is now the common ancestor of the two structs actually used). So, assuming I haven't missed anything, there is no user-visible change for test code that interacts with spy contexts for successful function calls.

snogge:
I had another read through after your doc-update, and of course I noticed some stuff :)

I didn't like the way `make-spy-context-return` was used, as I think that exposes implementation details.  So I tweaked the structs a bit to get a single constructor that should be semantically quivalent to what we had before, and also tried to improve the error messages. 

The error messages from the `return-value` and `thrown-signal` accessors can probably be improved further.

I pushed it to https://github.com/snogge/emacs-buttercup/commits/struct-tweaks, what do you think?

DarwinAwardWinner:
I like your implementation a lot better. It basically does what I wanted to do but didn't know how because I'm not an expert on `cl-defstruct`.

DarwinAwardWinner:
My one bit of feedback would be that I think you should require every call to `make-spy-context` to either provide a return value or thrown signal. I don't see the benefit in defaulting to `:return-value nil`.

snogge:
> My one bit of feedback would be that I think you should require every call to `make-spy-context` to either provide a return value or thrown signal. I don't see the benefit in defaulting to `:return-value nil`.

This was to ensure backward compatibility as that was valid before.

DarwinAwardWinner:
Ok, if that's how it worked before, then fair enough. Maybe we should deprecate that, but that's another issue. At the very least, I think the tests ought to demonstrate explicitly providing a return value even when it's nil.

snogge:
I added an expect that uses both `:return -value` and `:thrown-signal`. I will merge this PR now, and then my branch of tweaks.

DarwinAwardWinner:
Thanks for whipping this into shape!

PR #147, opened 2019-09-03
Do not leak functions from "The Spy " suite

snogge:
The (symbol-function test-function) cannot be lexically bound, but
equivalent behaviour can be acheived using before-all and after-all.

This is an alternative (better?) solution for 769e7b025549c5e64d011a0ecfe675b48c5eecab in #146. 

#

### DIFF
--- a/tests/test-buttercup.el
+++ b/tests/test-buttercup.el
@@ -613,16 +613,27 @@
 ;;; Spies
 
 (describe "The Spy "
-  (let (test-function)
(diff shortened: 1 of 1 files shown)


DarwinAwardWinner:
All looks good. You might consider renaming the variables to `saved-test-function` instead of `test-function` and so on, to reflect the new purpose of those variables.

snogge:
Finally got it to pass tests even on 24.3.

PR #152, opened 2019-12-20
Update test for new warning capture suppression behavior

DarwinAwardWinner:
While the test was technically correct, the text of the warning was inaccurate (it was based on a previous version of the test). In addition, running the tests multiple times within the same Emacs could give false positive results, since the warning text from the first test run would be included in the \*Warnings\* buffer. The new test incorporates the current time into the warning text, so as long as test runs are at least a second apart, there will be no false positives.


snogge:
The travis failures are all due to the bad CI environment fixed in #155.

PR #153, opened 2019-12-20
Don't overwrite pending status with passed status

### LINKED ISSUE #149: after-each overwrite pending state
I'm writing a test suite for ido. Using current buttercup, I get the following output (also visible on Travis [here](https://travis-ci.org/DarwinAwardWinner/ido-test-suite/jobs/607818075)):

```
$ cask exec buttercup -L included-ido tests
Ido location: "/home/travis/build/DarwinAwardWinner/ido-test-suite/included-ido/ido.el"
Running 17 out of 21 specs.
ido
  ido-completing-read
    should accept a matching completion (80.89ms)
    should complete with a matching item on RET (34.91ms)
    should complete with the first match when multiple matches are available (24.92ms)
    should allow <left> and <right> to cycle completions, with wrap-around (35.82ms)
    should allow C-j to exit with a partial completion (35.29ms)
  ido file operations
    ido-read-file-name
      Should allow selecting an existing file with RET (49.05ms)
      Should allow falling back to default completion with C-f (49.16ms)
    ido-read-directory-name
      Should allow selecting an existing directory with RET (48.64ms)
      Should not allow selecting an existing non-directory file with RET (35.25ms)
    ido-find-file
      should allow finding an existing file with RET (42.14ms)
      should allow finding new files (51.91ms)
    ido-write-file
      should allow writing new files with RET RET (37.55ms)
      should allow overwriting existing files with RET y (20.46ms)
    ido-dired
  ido buffer operations
    ido-read-buffer
      should allow selecting an existing buffer with RET (36.32ms)
      should allow selecting a new buffer name with RET RET (35.77ms)
      should allow falling back to default completion with C-f (43.21ms)
    ido-switch-buffer
      should allow switching to an existing buffer with RET (33.74ms)
      should allow creating a new buffer with RET RET (35.18ms)
  regression tests
    should not exhibit bug #19412 (19.45ms)
    should not exhibit bug #11861 (28.42ms)
    should not exhibit bug #1516 (20.58ms)
Ran 21 specs, 0 failed, in 0.8 seconds.
```

Buttercup correctly indicates at the top that several specs are not run, but at the bottom it just says "ran 21 specs", and every spec is output in green with timing info adjacent, which means I can't tell which specs were actually run.

Looking at the [code for this commit](https://github.com/DarwinAwardWinner/ido-test-suite/blob/091b0311946d5448629e7cc317d742ae4810859d/tests/test-ido.el#L474-L522), you can see that the last 3 tests are decalred with `xit`, which means they should be marked as PENDING.

DarwinAwardWinner:
Previously, if a pending spec was afftected by an after-all form,
Buttercup would evaluate the after-all form after not running the
pending test, and the status of the after-all form would be used as
the result of the test, marking it as passed instead of pending. The
logic in buttercup--update-with-funcall has been updated so that a
pending status will never be overwritten by a passed status.

I don't think there's an easy way to add an automated test for this
fix, since it is a fix in the core code itself. However, you can
manually test this by adding an after-all form in the "Pending Specs"
section of docs/writing-tests.md and verifying in the test output that
these tests are still marked as pending. (With the old code, they get
marked as passed instead.)

Fixes #149.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1423,20 +1423,47 @@ Sets the `status', `failure-description', and `failure-stack' for
 failed and pending specs."
   (let* ((result (apply 'buttercup--funcall function args))
          (status (elt result 0))
+         (prev-status (buttercup-suite-or-spec-status suite-or-spec))
          (description (elt result 1))
          (stack (elt result 2)))
-    (when (eq status 'failed)
-      (pcase description
-        (`(error (buttercup-failed . ,failure-description))
-         (setq description failure-description))
-        (`(error (buttercup-pending . ,pending-description))
-         (setq status 'pending
-               description pending-description))))
-    (when (memq (buttercup-suite-or-spec-status suite-or-spec)
-                '(passed pending))
-      (setf (buttercup-suite-or-spec-status suite-or-spec) status
-            (buttercup-suite-or-spec-failure-description suite-or-spec) description
-            (buttercup-suite-or-spec-failure-stack suite-or-spec) stack))))
+    (when (not (eq prev-status 'failed))
+      (when (eq status 'failed)
+        (pcase description
+          (`(error (buttercup-failed . ,failure-description))
+           (setq description failure-description))
+          (`(error (buttercup-pending . ,pending-description))
+           (setq status 'pending
+                 description pending-description))))
+      ;; Failed overrides pending, which overrides passed. Note that
+      ;; `prev-status' being failed is already covered above
+      (when (or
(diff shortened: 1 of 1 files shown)


snogge:
I already had the fix done when you posted yours, and unless you have objections I think I will merge #156 instead of this PR. It's a less invasive fix with fewer special cases. 

The test may be a bit specific, but I'm not sure how I could make it more general.

snogge:
Another fix for #149 has been merged, closing this PR.

PR #154, opened 2019-12-20
Add emacs-26.3 to CI config

DarwinAwardWinner:
Time to add Emacs 26.3 to Travis.


snogge:
There is a problem with the container image we're using. I tried to fix it here https://github.com/snogge/emacs-buttercup/commit/6e7e31e98be3d54bdeea59a6fcc161f9fe510c80 but it didn't work.

Issue #30, opened 2016-01-10
Spec count in output does not account for skipped specs


lunaryorn:
Buttercup doesn't exclude skipped specs from the number of running specs.  Consider the following output from our Flycheck specs.  Note how only four specs actually run, the rest is skipped, but buttercup claims that it's running all 36 specs, and afterwards says that it ran all 36 specs.

```
$ cask exec buttercup -L . --pattern ^Language Haskell
c-l-a-l: ("--pattern" "^Language Haskell")
Running 36 specs.

Language Haskell
  Module names
    does not extract a module name from commented code
    extracts a simple module name without exports
    extracts a simple module name at the end of a line
    extracts a module name with exports

Error List
  has the correct buffer name  SKIPPED
  has a permanently local source buffer  SKIPPED
  derives from Tabulated List Mode  SKIPPED
  Format
    sets the error list format locally  SKIPPED
    sets a proper padding locally  SKIPPED
    sets the list entries locally  SKIPPED
    has a local header line  SKIPPED
  Columns
    has the line number in the 1st column  SKIPPED
    has the column number in the 2nd column  SKIPPED
    has the error level in the 3rd column  SKIPPED
    has the error ID in the 4th column  SKIPPED
    has the error message in the 5th column  SKIPPED
  Entry
    has the error object as ID  SKIPPED
    has the line number in the 1st cell  SKIPPED
    has the column number in the 2nd cell  SKIPPED
    has an empty 2nd cell if there is no column number  SKIPPED
    has the error level in the 3rd cell  SKIPPED
    has the error ID in the 4th cell  SKIPPED
    has the error message in the 5th cell  SKIPPED
    has a default message in the 5th cell if there is no message  SKIPPED
  Filter
    kills the filter variable when resetting the filter  SKIPPED
    filters errors with lower levels  SKIPPED
    Mode Line
      shows no mode line indicator if no filter is set  SKIPPED
      shows the level filter in the mode line if set  SKIPPED

Manual
  List of languages
    lists all languages  SKIPPED
    doesn't list unknown languages  SKIPPED
  Syntax checkers
    documents all syntax checkers  SKIPPED
    doesn't document syntax checkers that don't exist  SKIPPED
    Options
      documents all syntax checker options  SKIPPED
      doesn't document syntax checker options that don't exist  SKIPPED
    Configuration files
      documents all syntax checker configuration files  SKIPPED
      it doesn't document configuration files that don't exist  SKIPPED

[...]

PR #158, opened 2019-12-27
Exclude skipped tests from first report line

snogge:
The first line in a report:

```
Running 113 specs.
```

did take pending tests into account, but not tests that were skipped
by the `bin/buttercup` `-p` option.  Setting the spec `status` field
to `pending` will exclude the skipped tests from the "Running" count. 

----

#

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1368,14 +1368,19 @@ current directory."
         (when (not (string-match "\\(^\\|/\\)\\." (file-relative-name file)))
           (load file nil t))))
     (when patterns
-      (dolist (spec (buttercup--specs buttercup-suites))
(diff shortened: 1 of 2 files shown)


snogge:
Also see #30.

doublep:
Please consider renaming `buttercup--mark-skipped` to `buttercup-mark-skipped` so that it becomes part of official API and can be used by other packages without fear of being removed in a next version.

I would also separate this into two functions: one that marks specs skipped according to some predicate and one that calls the first after building a predicate from a list of patterns. This way existing behavior of patterns would be preserved (the second function does it), while the first provides a very generic way to select using an arbitrary lambda.

Issue #161, opened 2020-01-16
Need option to not print skipped tests


doublep:
I'm the author of [Eldev build tool](https://github.com/doublep/eldev) and want to add integration with Buttercup library. Currently, Eldev supports only ERT as far as testing goes. Ideally, I want to make Buttercup testing interface as close to what I have for ERT as possible, so that testing framework affects how you write the tests, but not so much how you run them.

However, I stumble at the very first step. In ERT there are selectors. They provide a more-or-less simple way to pick *which* tests you want run. For that matter, testing libraries I use for other languages (e.g. JUnit in Java) provide similar facilities. But I don't see anything comparable in Buttercup. Is the only way to "select" tests in Buttercup to avoid loading certain files or disable unwanted tests with `xit`?

jorgenschaefer:
Hm. Buttercup is _mainly_ meant to be used non-interactively, so you might stumble on some issues.

There is `buttercup-run-at-point`, and `buttercup-run-discover` supports a pattern argument, but there is no direct API function you can use.

doublep:
> meant to be used non-interactively

There might be some misunderstanding here.  Eldev is also a non-interactive tool, it is run from command-line (of course you can use e.g. Eshell for this, but in any case it is disconnected from your "normal" Emacs).  For example, with ERT you can do:

    $ eldev test foo

and it will run all the tests with `foo` in the name and give you results — all in the terminal.  One of the purposes, of course, is to run on a continuos integration server, where you actually want to run all the tests anyway.  But it is also useful during normal development in cases where you don't plan to *interactively* debug test failures.  For example, because it shields you from all other miscellaneous packages in your normal Emacs, incomplete and obsolete evaluations and so on.

> `buttercup-run-discover `

This looks promising, thank you.  In particular, the patterns look somewhat like ERT selectors.  In Eldev I'd need to parallel this function inside the tool, but this itself should be no problem.

Two questions.  In Buttercup project itself this function fails with a misterious error (maybe there are supposed to be additional setup steps, but I couldn't figure out from `README.md`):

    ~/git/buttercup$ emacs -batch -f package-initialize -f buttercup-run-discover
    Cannot open load file: No such file or directory, ./tests/test-buttercup.el

And is there any compelling reason function `buttercup--specs` is package-private, i.e. not called `buttercup-specs` instead?

Thank you.

jorgenschaefer:
[...]

PR #162, opened 2020-01-22
Optionally don't print anything about skipped tests

doublep:
Here is an implementation of the feature of "don't print anything about skipped tests" I proposed in issue #161. The diff might look somewhat big, but it is only addition of new code, there are no removals. This can be verified by rediffing it with whitespace ignored: all "remove" hunks are only because of reindenting.

Everything revolves around new variable `buttercup-silent-skipping`. I initialized it to nil to preserve current behavior, but personally think default value of t would be more logical.

When this variable is non-nil, all output from Buttercup is put on hold and kept in the list `buttercup--pending-output`. This is done because when we discover that a test is skipped, we need to "erase" already generated output, possibly going as far as not mentioning its suite at all (if everything in the suite is skipped). Other implementations are possible by not running anything skipped at all or changing order of events, introducing new events, etc., but this all would break backward compatibility. Proposed implementation only alters the built-in reporters, not touching the way tests are executed at all.

External reporters will not get skipping functionality for free, but they can implement something similar based on new function `buttercup-spec-omitted-p`.

I added two tests for new functions. Can write more if needed.

New functionality is available from command line too. Here is an example:

    $ ./bin/buttercup -L . -p functions -s
    Running 117 specs.
    
    The `buttercup--apply-matcher' function
      should work with functions (34.13ms)
    
    The Spy 
      `spy-on' function
        can spy on autoloaded functions (14.85ms)
    
    Ran 2 out of 117 specs, 0 failed, in 0.1 seconds.
    paul@gonzo:~/git/buttercup$

### DIFF
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -31,6 +31,11 @@ Buttercup options:
                           which case tests will be run if they match
                           any of the given patterns.
 
+--silent-skipping, -s   When a test is skipped due to patterns or
+                          `xit' form, don't print anything about it.
+                          Likewise, don't omit all suite output if all
+                          specs inside it are skipped.
+
 --no-color, -c          Do not colorize test output.
 
 --traceback STYLE       When printing backtraces for errors that occur
@@ -70,7 +75,7 @@ do
             shift
             shift
             ;;
-        "-c"|"--no-color")
+        "-c"|"--no-color"|"-s"|"--silent-skipping")
             BUTTERCUP_ARGS+=("$1")
             shift
             ;;
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -32,7 +32,7 @@ Buttercup options:
                           any of the given patterns.
 
 --silent-skipping, -s   When a test is skipped due to patterns or
-                          `xit' form, don't print anything about it.
+                          "xit" form, don't print anything about it.
                           Likewise, don't omit all suite output if all
                           specs inside it are skipped.
 
--- a/buttercup.el
+++ b/buttercup.el
@@ -1539,6 +1539,11 @@ EVENT and ARG are described in `buttercup-reporter'."
 (defvar buttercup-reporter-batch--failures nil
   "List of failed specs of the current batch report.")
 
+(defvar buttercup--pending-output t
+  "List of pending output strings, reversed.
+Can also be symbol t to print any output immediately. The list
+can contain nil items that serve as level separators.")
+
 (defun buttercup-reporter-batch (event arg)
   "A reporter that handles batch sessions.
 
@@ -1729,11 +1734,6 @@ EVENT and ARG are described in `buttercup-reporter'."
      ;; Fall through to buttercup-reporter-batch implementation.
      (buttercup-reporter-batch event arg)))
   )
-
-(defvar buttercup--pending-output t
-  "List of pending output strings, reversed.
-Can also be symbol t to print any output immediately. The list
-can contain nil items that serve as level separators.")
 
 (defun buttercup--print (fmt &rest args)
   "Format a string and send it to terminal without alteration.
--- a/buttercup.el
+++ b/buttercup.el
@@ -663,6 +663,9 @@ See also `buttercup-define-matcher'."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Suite and spec data structures
 
+(defvar buttercup-silent-skipping nil
+  "Don't print anything about skipped tests.")
+
 (cl-defstruct buttercup-suite-or-spec
   ;; The name of this specific suite
   description
@@ -757,6 +760,12 @@ See also `buttercup-define-matcher'."
                 " "
                 (buttercup-spec-description spec))
       (buttercup-spec-description spec))))
+
+(defun buttercup-spec-omitted-p (spec)
+  "Determine if the SPEC should be omitted from the output."
+  (and buttercup-silent-skipping
+       (eq (buttercup-spec-status spec) 'pending)
+       (equal (buttercup-spec-failure-description spec) "SKIPPED")))
 
 (defun buttercup--full-spec-names (spec-or-suite-list)
   "Return full names of all specs in SPEC-OR-SUITE-LIST."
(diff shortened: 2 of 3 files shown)


bin/buttercup
```diff
@@ -31,6 +31,11 @@ Buttercup options:
                           which case tests will be run if they match
                           any of the given patterns.
 
+--silent-skipping, -s   When a test is skipped due to patterns or
+                          `xit' form, don't print anything about it.
```
snogge: That quoting style wont work in this type of heredoc. You need to switch to ""

buttercup.el
```diff
@@ -1536,46 +1548,55 @@ EVENT and ARG are described in `buttercup-reporter'."
     (pcase event
       (`buttercup-started
        (setq buttercup-reporter-batch--start-time (float-time)
-             buttercup-reporter-batch--failures nil)
+             buttercup-reporter-batch--failures nil
+             ;; Don't put any output on hold if nothing is ever omitted.
+             buttercup--pending-output (not buttercup-silent-skipping))
```
snogge: buttercup--pending-output should be defined before this defun.

buttercup.el
```diff
@@ -1536,46 +1548,55 @@ EVENT and ARG are described in `buttercup-reporter'."
...
-                       (list arg))))
-        ((eq (buttercup-spec-status arg) 'pending)
-         (buttercup--print "  %s" (buttercup-spec-failure-description arg)))
-        (t
-         (error "Unknown spec status %s" (buttercup-spec-status arg))))
-       (buttercup--print " (%s)\n"
-                         (seconds-to-string
-                          (float-time (buttercup-elapsed-time arg)))))
+       (if (buttercup-spec-omitted-p arg)
+           (buttercup--discard-one-output-level)
```
snogge: It should be possible to move this into the cond as the first case. Then you wouldn't need any reindentation of the rest of the code.

buttercup.el
```diff
@@ -1536,46 +1548,55 @@ EVENT and ARG are described in `buttercup-reporter'."
...
+                           (seconds-to-string
+                            (float-time (buttercup-elapsed-time arg))))
+         (buttercup--flush-pending-output)))
 
       (`suite-done
-       (when (= 0 (length (buttercup-suite-or-spec-parents arg)))
-         (buttercup--print "\n")))
+       (unless (buttercup--discard-one-output-level)
+         (when (= 0 (length (buttercup-suite-or-spec-parents arg)))
+           (buttercup--print "\n"))))
```
snogge: (or (buttercup--discard-one-output-level)
     (< 0 (length ....
     (buttercup--print  ...
The `=` sexp can be replaced with `(null (buttercup-suite-or-spec-parent arg))`.

buttercup.el
```diff
@@ -1627,31 +1649,34 @@ colors.
...
-                           (buttercup-spec-description arg))
-         (setq buttercup-reporter-batch--failures
-               (append buttercup-reporter-batch--failures
-                       (list arg))))
-        ((eq (buttercup-spec-status arg) 'pending)
-         (if (equal (buttercup-spec-failure-description arg) "SKIPPED")
-             (buttercup--print "  %s" (buttercup-spec-failure-description arg))
-           (buttercup--print (buttercup-colorize "\r%s%s  %s" 'yellow)
+     (if (buttercup-spec-omitted-p arg)
+         (buttercup--discard-one-output-level)
```
snogge: Move into the cond.

tests/test-buttercup.el
```diff
@@ -1107,11 +1109,32 @@
...
+      (expect buttercup--pending-output :to-equal '("not to be printed just yet"))))
+
+  (it "should discard one level of output"
+    (let ((buttercup--pending-output nil))
+      (buttercup--print "1")
+      (buttercup--start-output-level)
+      (buttercup--print "2")
+      (buttercup--discard-one-output-level)
+      (buttercup--print "3")
+      (buttercup--flush-pending-output))
```
snogge: Add an expect eher to check that send-string-to-terminal has not been called yet.

bin/buttercup
doublep: Fixed.

buttercup.el
doublep: Fixed.

buttercup.el
doublep: No, it's not possible because the structure is `(if X (Y) (cond ...) (Z))`. I.e. all `cond` clauses lead to `(Z)` branch, but `(Y)` must not.

buttercup.el
doublep: It's a matter of style, I thought it would be more explicit this way.

buttercup.el
doublep: Not possible, see above.

tests/test-buttercup.el
doublep: Done.

doublep:
I will squash the new commit later, wanted to leave it visible for verification.

> What do you think of #164 in comparison?

Looks fine for the problem in question. However, my approach would be extendable by altering `buttercup-spec-omitted-p`. You also have a function `buttercup-omitted-p`, but it is called *before* suite/spec is executed, restricting its options heavily.

What I specifically had in mind is adding another option and defvar `buttercup-quiet`. If used, only suites/specs that resulted in errors would be printed. This naturally can only be decided after running the tests and so your solution cannot be extended to support this, as far as I can see.

If idea of `buttercup-quiet` is rejected completely (even for future), then either solution is fine to me.

doublep:
I've just had another idea: instead of putting output on hold, how about putting *events* on hold?

Outline: create a `(defun buttercup-reporter-delaying-adapter (base-reporter) ...)`. It would return a reporter based on given one that would put events on hold and potentially discard them if not interesting according to "silent skipping" or "quiet" setting. If they become interesting (a test is not skipped / a test fails), event from the pending queue gets sent to the `base-reporter`.

Advantages:

* Over this (#162): easily reusable by other (external) reporters: just use `(buttercup-reporter-delaying-adapter 'my-cool-reporter)`. But you still have to opt-in, so we don't break compatibility.

* Over #164: more open to extending, e.g. for `buttercup-quiet`.

Disadvantages:

* Reporters that do opt-in miss certain events completely as they get discarded. However, we can also convert them to sth. like `SIGNAL-NAME:omitted` so opting-in reporters still get them and may process them somehow should they want. This still wouldn't break compatibility, because such additional signals would only get sent to opting-in reporters, i.e. those aware of new functionality.

doublep:
I'm closing this as #166 is certainly a better approach.

PR #164, opened 2020-01-23
Do not report skipped specs

snogge:
When `buttercup-silent-skipping` is non-nil, do not print specs that
were marked as skipped before `buttercup-run` was started.  Also do
not print the descriptions for suites that only contain such skipped
specs.

Specs that are marked as skipped by `assume` will still be printed.

Add an option `--silent-skipping` to bin/buttercup.

### DIFF
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -31,6 +31,11 @@ Buttercup options:
                           which case tests will be run if they match
                           any of the given patterns.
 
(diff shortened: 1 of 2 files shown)


snogge:
This is an alternative implementation to #162 that still prints the
output as each spec is processed.  It does have the drawback that it
will still print specs that are skipped by raising the
`buttercup-pending` signal.

snogge:
A better solution was merged from #184

PR #166, opened 2020-01-24
Optionally don't print anything about skipped and/or passed tests

doublep:
Yet another implementation of the feature of "don't print anything about skipped tests" from issue #161. Additionally this implements one more option `-q`. When that is set, Buttercup prints *only* failed tests (with `-s` it prints failed and passed, by default it prints all, including skipped, as now).

This implementation adds function `buttercup-reporter-delaying-adapter` that can sit in front of any reporter and filter events sent to it according to variables `buttercup-silent-skipping` and `buttercup-quiet`. Unlike other proposed implementations, this can be used for *any* reporter by adding a single call to the adapter.

The implementation is also fairly small and almost doesn't touch existing code. It's quite unobvious, but I didn't add any comments since I felt like I would comment every single line then.

I failed to write tests (only tested manually) for this implementation, some commented-out attempt is in the commit. @snogge: can you help with this?

### DIFF
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -31,6 +31,15 @@ Buttercup options:
                           which case tests will be run if they match
                           any of the given patterns.
 
+--silent-skipping, -s   When a test is skipped due to patterns or
+                          "xit" form, don't print anything about it.
+                          Likewise, don't omit all suite output if all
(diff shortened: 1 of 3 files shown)


snogge:
Another solution was merged from #184

Issue #157, opened 2019-12-26
spec *-each and function should only set "more serious" state


snogge:
As discussed in #149, any `before-each`, `buttercup-spec-function` or `after-each` should only set the state of the spec if the new state is "more serious" than the specs current state. 

This means that `pending`should win over `success` and `failure` should win over both `pending` and `success`.  `success` should never win over any other state. 

All in all we think this will improve the test results.

DarwinAwardWinner:
I'm not sure if this is the same bug, but using `assume` in a `before-each` form skips the remainder of the `before-each` but then seems to run the body of the test as normal (which then fails because the assumption was false).

In addition, using `assume` in the body of a test gets overwritten with a passed state if there is an `after-each` form. I'm pretty sure this example *is* an instance of this bug.

Also, I don't believe the `assume` macro is documented anywhere. I'm not even certain whether it's meant to be used in a `before-each` form, but it seems like it should work there.

PR #167, opened 2020-01-27
Allow clean abort

doublep:
Would Buttercup be interested in the following functionality?

1. Test run can be aborted cleanly at any point (by a test, by the reporter). "Cleanly" means that corresponding `*-done` for started specs and suites are emitted, the run ends in `buttercup-done` event, but no more specs are executed after aborting.

2. New option: immediately abort if a test fails. This is useful if your tests are so interdependent (at least in the context of the current bug) that one failure typically means there will be 20 more. Also when tests are slow enough that you don't want to wait for all of them to finish to get the backtrace of the first failure.

### DIFF
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -31,6 +31,8 @@ Buttercup options:
                           which case tests will be run if they match
                           any of the given patterns.
 
+--abort-on-failure, -a  Abort test execution if any test fails.
+
 --no-color, -c          Do not colorize test output.
 
 --traceback STYLE       When printing backtraces for errors that occur
@@ -70,7 +72,7 @@ do
             shift
             shift
             ;;
-        "-c"|"--no-color")
+        "-c"|"--no-color"|"-a"|"--abort-on-failure")
             BUTTERCUP_ARGS+=("$1")
             shift
             ;;
--- a/buttercup.el
+++ b/buttercup.el
@@ -671,6 +671,9 @@ See also `buttercup-define-matcher'."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Suite and spec data structures
 
+(defvar buttercup-stop-on-first-failure nil
+  "Abort Buttercup testing if any spec fails.")
+
 (cl-defstruct buttercup-suite-or-spec
   ;; The name of this specific suite
   description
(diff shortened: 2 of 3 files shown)


jorgenschaefer:
Thank you for your contribution! :-)

snogge:
I'm not saying we should not have this functionality, but I would like to revert it now, refine it a bit and possibly merge it later. Here's why:

* There are no tests
* There is no documentation
* It potentially breaks other reporters, I know butterup-junit will generate invalid reports if this is activated. The correct thing is probably to mark all remaining specs and suites as pending and run them as all other pending specs.  This ties into #149 and #157. If we combine this PR with #161/#166 we get more or less the behaviour that this PR aims for.
* If the failing spec has `after-each` or `after-all` functions these are not run.
* The spec counts at the end of the report will be wrong, any spec that were pre-empted will be reported as if they had been run, while they were actually skipped.

I think @doublep meant this PR mostly as an example of what he wanted to achieve anyway.
What do you think, @jorgenschaefer and @doublep?

By the way, I added these tests to check the reporter and the before/after functions: 
```emacs-lisp
(describe "When tests abort"
  (it "the before and after functions should still run"
    (let (all-before all-after each-before each-after)
      (expect
       (with-local-buttercup
         (setq buttercup-stop-on-first-failure t)
         (describe "suite"
           (before-all (setq all-before t))
           (after-all    (setq all-after t))
           (before-each 
             (setq each-before t)
             (spy-on 'not-a-real-function))
           (after-each (setq each-after t))
           (it "spec that fails"
             (not-a-real-function)
             (expect nil)))
         (buttercup-run)) :to-throw)
      (expect (symbol-function 'not-a-real-function) :to-be nil)
      (expect (list all-before all-after each-before each-after) :to-equal '(t t t t)))))

(describe "The `buttercup-run' function"
  ;; .....
  (it "should call all reporter steps even on abort"
    (let (reporter-args)
      (with-local-buttercup
       (setq buttercup-reporter (lambda (&rest args) (push args reporter-args)))
       (describe "suite"
         (it "spec"
           (signal 'buttercup-abort "buttercup-abort")))
       (buttercup-run))
      (expect (nreverse (mapcar #'car reporter-args)) :to-equal
              '(buttercup-started
                suite-started
                spec-started
                spec-done
                suite-done
                buttercup-done))))
)
```

jorgenschaefer:
Sorry, you are right.

snogge:
I have reverted the two commits from this PR that @jorgenschaefer cherry-picked to master.  I don't know whether this PR can be reopened, but otherwise feel free to open a new PR.  I've mentioned some things above that I think need to be addressed before this PR can be merged.  My focus right now is finding solutions for #149 and #157, but I will return to this and @doublep's other [eldev](https://github.com/doublep/eldev)-related PRs eventually.

PR #169, opened 2020-03-22
Fixed misspelling in the documentation

sergeyklay:
Hello,

There is a fix for few minor typos.


snogge:
Actually, I think @jorgenschaefer used `iff' in the mathematical sense, meaning `if and only if'.

jorgenschaefer:
I did, but that's super nerdy …

sergeyklay:
Got it. Should I close this then?

jorgenschaefer:
Maybe merge it to avoid further confusion? @snogge what do you think?

PR #172, opened 2020-05-02
Make sure carriage movement characters are not colorized

### LINKED ISSUE #171: Missing colors in compilation buffer
I would like to have the same output (or at least something close) in Emacs compilation buffers as I have when running buttercup from the shell: https://imgur.com/a/U9a3pzE

But as you can see the colors for the individual tests are missing. I use the following code to translate ansi colors:

```Elisp
(defun colorize-compilation ()
  (let ((inhibit-read-only t))
    (ansi-color-apply-on-region compilation-filter-start (point))))
(add-hook 'compilation-filter-hook #'colorize-compilation)
```
It does work in general since the `0 failed` part is correctly made green.

Also without this hook I only see escape sequences to reset the color, but not to set it:  https://imgur.com/a/x2qStz8, so I'm not sure whether this is a bug report or a feature request, but it'd be great if it was possible to make left and right look the same in these pictures.

snogge:
Fixes #171, which reports that adding `ansi-color-apply-on-region` to
`compilation-filter-hook` did not work with colorized buttercup output.
This was because in the `compilation-filter` function, carriage
movment characters `\r` and `\n` are handled (by
`comint-carriage-motion? ) before `compilation-filter-hook? .

It is often recommended to use
```emacs-lisp
;; either `point' or `point-max' work as the second argument
(ansi-color-apply-on-region compilation-filter-start (point))
```
in a `compilation-filter-hook` function to handle ANSI or SGR control
sequences in compilation output.  But carriage motion may cause
`point` to move to before `compilation-filter-start`.  Then
`ansi-color-apply-on-region` will miss SGR control sequences in the
`point` --- `compilation-filter-start` range.  Using
```emacs-lisp
(ansi-color-apply-on-region
    (save-excursion
        (goto-char compilation-filter-start)
        (line-beginning-position))
    (point))))
```
instead will at least avoid this problem for carriage motion that does
not move point to preceding lines.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1653,25 +1653,25 @@ EVENT and ARG are described in `buttercup-reporter'."
      (unless (string-match-p "[\n\v\f]" (buttercup-spec-description arg))
        (buttercup-reporter-batch event arg)))
     (`spec-done
-       (cond
-        ((eq (buttercup-spec-status arg) 'passed)
-         (buttercup--print (buttercup-colorize "\r%s" 'green)
-                           (buttercup--indented-description arg)))
-        ((eq (buttercup-spec-status arg) 'failed)
-         (buttercup--print (buttercup-colorize "\r%s  FAILED" 'red)
-                           (buttercup--indented-description arg))
-         (setq buttercup-reporter-batch--failures
-               (append buttercup-reporter-batch--failures
-                       (list arg))))
-        ((eq (buttercup-spec-status arg) 'pending)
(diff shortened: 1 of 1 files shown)


snogge:
On second though, the ANSI codes that are not interpreted probably depends on how the input happens to be processed by the compilation process and is passed to it's input filer compilation-filter. This commit has no way to guarantee that the codes are interpreted correcvtly, it just makes them available to compilation-filter which is then free to screw it up.

PR #177, opened 2020-07-03
Add API for better skip selection

snogge:
The new function `buttercup-mark-skipped` lets users mark specs as pending/skipped using either regular expressions or predicate functions.

@doublep , what do you think of this API?  Maybe we should supply more data to the predicate function, like the full description string?

### DIFF
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -29,7 +29,10 @@ Buttercup options:
 --pattern, -p PATTERN   Only run tests with names matching PATTERN.
                           This option can be used multiple times, in
                           which case tests will be run if they match
-                          any of the given patterns.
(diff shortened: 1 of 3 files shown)


doublep:
Should be fine.

>Maybe we should supply more data to the predicate function, like the full description string?

As long as it receives the spec object and API to access it is public and stable, I don't see a need for that.

snogge:
OK.  I'll leave this PR open for a few more days in case anyone else wants to chime in.
By the way, I have a working solution for the rest of #161, but it needs some more work and tests before its ready to put up.  Frankly, the reporter is getting a bit too complicated by now.  I really liked it when it was stateless....

PR #183, opened 2020-07-15
Optionally fail when loading old elc files

### LINKED ISSUE #142: Uses outdated .elc files
I've noticed that sometimes while developing and testing, my tests may fail because of an outdated `.elc` file that I haven't recompiled yet.  Buttercup warns about this, but the warning quickly scrolls off the screen as it runs the tests.

It would be helpful if there were a simple way to make Buttercup either refuse to run if the byte-compiled file is outdated, or byte-compile it before running the tests.

I could write some code to do this in my test file, but it seems like a common issue, so it would be nice if there were a standard way to solve it.

Thanks.

snogge:
The new --stale-file-error option can be used to activate a hook that
will signal an error if a stale .elc file is loaded.  Stale means
older than its originating .el file.

Fixes #142.

### DIFF
--- a/bin/buttercup
+++ b/bin/buttercup
@@ -56,6 +56,8 @@ Buttercup options:
                           frame with a lambda or M to indicate whether
                           it is a normal function call or a
                           macro/special form.
+
+--stale-file-error      Fail the test run if stale .elc files are loaded.
 EOF
 }
 
@@ -81,7 +83,7 @@ do
             shift
             shift
             ;;
-        "-c"|"--no-color"|"--no-skip"|"--only-error")
+        "-c"|"--no-color"|"--no-skip"|"--only-error"|"--stale-file-error")
             BUTTERCUP_ARGS+=("$1")
             shift
             ;;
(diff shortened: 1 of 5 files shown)


bin/buttercup
```diff
@@ -83,6 +85,10 @@ do
             shift
             shift
             ;;
+		"--stale-file-error")
```
chrahunt, line 88: The rest of the file uses spaces, but these lines are using tabs.

buttercup.el
```diff
@@ -1969,6 +1972,29 @@ With buttercup minor mode active the following is activated:
...
 
+(defun buttercup-error-when-loading-old-elc-files (elc-file)
+  "Raise an error when ELC-FILE is an elc-file and older than its el-file."
+  (when (string= (file-name-extension elc-file t) ".elc")
+    (let ((elfile (substring elc-file 0 -1))
+          elc-attributes el-attributes)
+      (when (file-exists-p elfile)
+        (setq elc-attributes (file-attributes elc-file))
+        (setq el-attributes (file-attributes elfile))
+        (if (time-less-p (file-attribute-modification-time elc-attributes)
```
chrahunt, line 1983: When do we use `when` vs `if`?

bin/buttercup
snogge: Thanks.

buttercup.el
snogge: We generally use when in buttercup, I think.  There is no hard rule.  I've recently used plain `and` and `or` when I felt it made the code easier to understand.

snogge:
Waiting for tests and some cleanup. I'm redoing the install/uninstall functions as a more minor-mode-like function that can be used to turn on or off or toggle the feature. I really expect it to only be used to turn it on so two functions felt like overkill.

Thank you for reviewing the PR.

PR #196, opened 2020-08-20
Fix compilation warning about undefined variable.

phst:
Dynamic variables always need to be defined before use.


snogge:
Thank you.

PR #197, opened 2020-08-29
New traceback style 'omit'

snogge:
Before adding the new style I wanted tests for backtraces, and boy did that open a can of worms.
The backtrace collection was broken, and has probably been for a while. Finding a good way of doing 
it was tricky, but I think most cases are handled in an OK way now.

The new omit style has to work a bit differently from the other styles, as not only the output of backtrace frames are omitted, but also headers and such things.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,4 +25,4 @@ release: clean test
 	tar -c $(DISTFILES) --transform "s,^,buttercup-$(VERSION)/," --transform 's/README.md/README.txt/' > "dist/buttercup-$(VERSION).tar"
 
 clean:
-	rm -f *.elc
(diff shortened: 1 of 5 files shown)


doublep:
If source files are not byte-compiled, there are 34 fails. Not sure this is a good idea: you make your tests depend on external tools. E.g. if run as `make test-buttercup`, everything works; if run as `./bin/buttercup -L . tests` (after `make clean` first if needed), there are lots of failures.

snogge:
Thank you for finding this.

snogge:
This passes the test both compiled and not. I've also run them on all the Emacs versions using the github actions.

PR #203, opened 2020-10-28
Make formatting of backtrace frames closer to that standard for Emacs

doublep:
With the following test:

    (it "fails"
        (let ((x (lambda () (backtrace) xxx)))
          (funcall x)))

Buttercup currently produces the following backtrace:

      (let ((x #'(lambda nil (backtrace) xxx))) (funcall x))
      (funcall (lambda nil (backtrace) xxx))
      ((lambda nil (backtrace) xxx))
    error: (void-variable xxx)

You can compare this with `(backtrace)` output (also in the test):

      backtrace()
      (lambda nil (backtrace) xxx)()
      funcall((lambda nil (backtrace) xxx))
      (let ((x #'(lambda nil (backtrace) xxx))) (funcall x))

You can see that standard Emacs backtrace prints `funcall` *outside* the parens. With the proposed changes so does Buttercup:

      (let ((x #'(lambda nil (backtrace) xxx))) (funcall x))
      funcall((lambda nil (backtrace) xxx))
      (lambda nil (backtrace) xxx)()
    error: (void-variable xxx)

There is no need to create additional tests for the change, it gets caught by the existing tests already. So, I modified those that started failing (i.e. they asserted previous formatting, I changed them to assert the new formatting that is closer to what Emacs itself does).

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1982,7 +1982,10 @@ If STYLE is nil, use `buttercup-stack-frame-style' or `crop'."
   (setq style (or style buttercup-stack-frame-style 'crop))
   (pcase style
     (`omit) ; needed to verify valid styles
-    (`full (format "  %S" (cdr frame)))
+    (`full
+     (if (car frame)
+         (format "  %S%s" (cadr frame) (if (cddr frame) (prin1-to-string (cddr frame)) "()"))
+       (format "  %S" (cdr frame))))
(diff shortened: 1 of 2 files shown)


snogge:
Thank you.

Too bad backtrace.el is only in Emacs 27.  Future improvement to use
it when available.

PR #204, opened 2020-11-22
Use an absolute path for the root of test files

akirak:
When I ran `buttercup-run-discover`, I encountered the following error:

> Cannot open load file: No such file or directory, ./org-multi-wiki-tests.el

It seems that `load` function sometimes (?) fails to load a file starting with `./`. For example, it fails when I run `buttercup-run-discover .` from the command line, while it succeeds when running `buttercup-run-discover $PWD`. I find it strange, but it would be safer to use `default-directory`.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1408,7 +1408,7 @@ current directory."
         (push (car args) dirs)
         (setq args (cdr args)))))
     (setq command-line-args-left nil)
-    (dolist (dir (or dirs '(".")))
+    (dolist (dir (or dirs (list default-directory)))
       (dolist (file (directory-files-recursively
                      dir "\\`test-.*\\.el\\'\\|-tests?\\.el\\'"))
         ;; Exclude any hidden directory, both immediate (^.) and nested (/.) subdirs


snogge:
Can you explain a bit more about this?  What was the command line you were using, and what is the file structure of your project?

akirak:
I ran the following command inside the repository of [org-multi-wiki](https://github.com/akirak/org-multi-wiki/):

```
emacs -Q --batch -l buttercup -f buttercup-run-discover
```

and received the following message:

> Cannot open load file: No such file or directory, ./org-multi-wiki-tests.el

On the contrary, I didn't get the error when I ran the following command:

```sh
emacs -Q --batch -l buttercup -f buttercup-run-discover /home/akirakomamura/.config/emacs/straight/repos/org-multi-wiki
```

akirak:
I ran `pwd; ls` before the command and apparently the file exists in the directory:

> /home/akirakomamura/.config/emacs/straight/repos/org-multi-wiki
> LICENSE  README.org  helm-org-multi-wiki.el  notes.org	org-multi-wiki-tests.el  org-multi-wiki.el  screenshots

akirak:
This seems to be irrelevant, so I will close it for now.

PR #215, opened 2022-03-21
Avoid a compilation warning.

phst:
Currently, compilation prints a warning:

    In toplevel form:
    buttercup.el:1933:1: Warning: variable ‘_data’ not left unused

The warning is incorrect (see
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26960), but I think it’s better
to avoid the warning so that the file compiles cleanly.


snogge:
Thank you.

PR #216, opened 2022-03-31
Collect backtraces for failed test file loads and display them as failed specs

### LINKED ISSUE #107: Errors in test files are silently ignored
If a test file throws an error, the error is ignored and any test suites that were defined before the error was thrown are run. Any test suites defined after the error are never defined or run, and buttercup erroneously reports a successful result even though some tests never ran. This problem masked some test failures is my expect-closures branch, as described here: https://github.com/jorgenschaefer/emacs-buttercup/pull/100#issuecomment-325148055.

snogge:
Use the existing error catching and backtrace collecting infrastructure when loading test files.  
For each test file that signals an error while loading, add a pre-failed spec to a special suite, and add that suite to the end of `buttercup-suites`.  These load failures will be listed at the end of the test report.
buttercup will still process any fully loaded specs, including spec from a partially loaded test file. 
The total buttercup run will fail as there are failing specs, and failure will be included in any test report even if the reporter does not print to stdout, like `buttercup-junit`.

Fixes #107

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1375,7 +1375,8 @@ current directory."
   (setq backtrace-on-error-noninteractive nil)
   (let ((dirs nil)
         (patterns nil)
-        (args command-line-args-left))
+        (args command-line-args-left)
+        failed-files-suite)
     (while args
       (cond
        ((equal (car args) "--")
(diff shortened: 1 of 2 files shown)


wasamasa:
This looks very nice. No apparent issues with the code, I'd only need to test whether it works as expected.

DarwinAwardWinner:
The approach seems reasonable to me. Perhaps the load of each file should unconditionally be captured in a suite? That way you'd get a report of every test file that was loaded along with whether the loading was successful. If you go with this idea, I would probably put this at the start of the report.

snogge:
I considered putting the failed loads at the start of the report, but felt they were more visible at the end. I guess it is a matter of opinion, taste and expectation. The failed loads happen before any specs are done, so it could be natural to put them first.
I'm not sure about making the file loads into specs.  When loads fail the spec count is probably messed up anyway, so adding specs to the report doesn't really matter.  But having the actual files used be part of the report could definitely be useful, maybe as a container of suites or other type of separate entry.

snogge:
I guess enough time has passed now.  I will merge this and deal with
any fallout later.

alphapapa:
@snogge Thanks for fixing this.

PR #231, opened 2023-02-27
docs(exec): Document use with Eask

jcs090218:
Hi, I want to add document regarding the use with Eask.

Cask is basically unmaintained, so I would like to introduce a new tool [Eask](https://github.com/emacs-eask/cli). It's very similar to Cask, with only a few syntaxes and command differences. LMKWYT?

Thanks!

### DIFF
--- a/docs/running-tests.md
+++ b/docs/running-tests.md
@@ -165,9 +165,110 @@ $ eldev test -f main.el
 For more information, please see tool’s own documentation, especially
 [the section about testing](https://github.com/doublep/eldev#testing).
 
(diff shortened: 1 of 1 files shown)


jcs090218:
cc @snogge

snogge:
I've seen it, haven't had time to review yet.

jcs090218:
Ah, okay. Sorry for the annoyance.

snogge:
Thank you.

jcs090218:
Thank you so much! :D

PR #232, opened 2023-03-04
Always set native-comp-enable-subr-trampolines to nil before running tests

### LINKED ISSUE #230: Native compilation problem when spying on certain primitives, such as file-exists-p
I'm investigating https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1028725, which is a bug currently affecting Debian's and Ubuntu's flycheck packages, and noticed a strange behaviour when two `spy-on`'s are used sequentially but the first one has an `:and-return-value` parameter.

The following reproducer always triggers the bug for me on Debian sid with an upstream Emacs (`Development version b26d0dd58371 on master branch; build date 2023-02-26`):

``` elisp
(describe "Test buffer-file-name"
  (it "test"
    (spy-on 'file-exists-p :and-return-value t)
    (spy-on 'buffer-file-name)))
```

Here's what I see:

```
$ buttercup --eval "(package-initialize)" -L .
Running 1 specs.

Test buffer-file-name
  test  FAILED (11.68ms)

========================================
Test buffer-file-name test

  spy-on(buffer-file-name)
  buttercup--spy-on-and-call-replacement(buffer-file-name (lambda (&rest arg...
  comp-subr-trampoline-install(buffer-file-name)
  comp-trampoline-search(buffer-file-name)
  native-elisp-load("/home/sergio/.emacs.d/eln-cache/30.0.50-2895a91e/subr--...
error: (native-lisp-load-failed "file does not exists" "/home/sergio/.emacs.d/eln-cache/30.0.50-2895a91e/subr--trampoline-6275666665722d66696c652d6e616d65_buffer_file_name_0.eln")

Ran 1 specs, 1 failed, in 24.59ms.
buttercup-run failed: ""
```

I'm not entirely sure whether this is a problem with buttercup or with Emacs' recent native compilation feature, but I thought I'd report this here in case you have any ideas.  Thanks.

sergiodj:
For more information, see
https://lists.gnu.org/archive/html/bug-gnu-emacs/2023-02/msg02464.html.

Closes #230

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1586,10 +1586,12 @@ Signal an error if any spec fail or if no suites have been
 defined. Signal no errors if NOERROR is non-nil. Return t if all
 specs pass, nil if at least one spec fail, and :no-suites if no suites
 have been defined."
-  (if buttercup-suites
-      (buttercup--run-suites buttercup-suites noerror)
-    (or (and noerror :no-suites)
-        (error "No suites defined"))))
+  ;; See https://github.com/jorgenschaefer/emacs-buttercup/issues/230.
+  (let ((native-comp-enable-subr-trampolines nil))
+    (if buttercup-suites
+        (buttercup--run-suites buttercup-suites noerror)
+      (or (and noerror :no-suites)
+          (error "No suites defined")))))
 
 (define-error 'buttercup-run-specs-failed "buttercup-run failed" 'buttercup-error-base)
 


snogge:
The threading in the archive is broken, see
https://lists.gnu.org/archive/html/bug-gnu-emacs/2023-03/msg00023.html
for the rest(?) of the thread

snogge:
I wont pretend I fully understand how the trampolines work or when
they are needed, but looking at the docs for
`native-comp-enable-subr-trampoline`

    native-comp-enable-subr-trampolines is a variable defined in ‘src/comp.c’.

    Its value is t

    If non-nil, enable generation of trampolines for calling primitives.
    Trampolines are needed so that Emacs respects redefinition or advice of
    primitive functions when they are called from Lisp code natively-compiled
    at ‘native-comp-speed’ of 2.

    By default, the value is t, and when Emacs sees a redefined or advised
    primitive called from natively-compiled Lisp, it generates a trampoline
    for it on-the-fly.

    If the value is a file name (a string), it specifies the directory in
    which to deposit the generated trampolines, overriding the directories
    in ‘native-comp-eln-load-path’.

    When this variable is nil, generation of trampolines is disabled.

    Disabling the generation of trampolines, when a trampoline for a redefined
    or advised primitive is not already available from previous compilations,
    means that such redefinition or advice will not have effect when calling
    primitives from natively-compiled Lisp code.  That is, calls to primitives
    without existing trampolines from natively-compiled Lisp will behave as if
    the primitive was called directly from C, and will ignore its redefinition
    and advice.

I'm a bit concerned about the last paragraph.  

I assume that this bug is triggered by the buttercup spy definitions,
which redefine functions and possibly primitives locally.  What if
that primitive does not already have a generated trampoline?  I'm
guessing the spy will be broken/ignored?

snogge:
I narrowed the scope where `native-comp-enable-subr-trampolines` is set to nil and added a test based on the error report. 
@sergiodj , thank you again for figuring this out.

sergiodj:
Sorry for the absence, @snogge; I got sidetracked with a bunch of things from work.

I see you've merged the change (thanks!), but regarding your initial question: during all my tests here, if you set `native-comp-enable-subr-trampolines` to `nil` then Emacs will just not try to compile the primitive into a trampoline, and will default to using the Elisp implementation instead.  I could not find a scenario where disabling the native compilation of primitives would lead to a failure in buttercup.

I will try to keep an eye on any possible side-effect of this change, and if needed we can think of a better way to cope with the situation.  Cheers.

PR #234, opened 2023-03-14
Add spy-arg-matcher for more complex matching of spy arguments

Fuco1:
More matchers to add for inspiration for example from Mockery: http://docs.mockery.io/en/latest/reference/argument_validation.html?highlight=matchers#additional-argument-matchers

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -701,7 +701,9 @@ UNEVALUATED-EXPR if it did not raise any signal."
      ((not calls)
       (cons nil
             (format "Expected `%s' to have been called with %S, but it was not called at all" spy args)))
-     ((not (member args calls))
+     ((not (cl-some
+            (lambda (call) (spy--match-args-to-call args call))
+            calls))
       (cons nil
             (format "Expected `%s' to have been called with %S, but it was called with %s"
                     spy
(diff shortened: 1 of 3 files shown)


snogge:
I like the idea, but I don't like the syntax. It would be cool if we could use `pcase` patterns for this.  That would probably require a new matcher that receives the argument patterns in a list.  The hardest part is probably to figure out a good name for the new matcher :smiley:

Fuco1:
I would argue against pcase or at least that we provide other syntax.  Many people including myself find the pcase syntax quite complicated to read.

>  That would probably require a new matcher that receives the argument patterns in a list.

Something like adding `:args-to-match-pattern` ? I kept the existing matcher for "backward" compatibility so it's completely opt-in.  A new matcher would also achieve that of course.

snogge:
The current `:to-have-been-called-with` expects the arguments as &rest:
```elisp
(expect 'foo :to-have-been-called-with 1 2 3)
```
That makes it difficult to make a pattern matching matcher, how do I see that 
```elisp
(expect 'foo :to-have-been-called-with 1 'predfunc 3)
```
is supposed to use `predfunc` as a predicate instead of expecting that symbol as an argument?  Hard to change in a backward compatible way.
A new matcher could use
```elisp
(expect 'foo :to-have-been-called-matching '(1 2 3))
(expect 'foo :to-have-been-called-matching 'predfunc)
```
and easily see that `predfunc` shall be used as a matching predicate.  If the single argument to `:to-have-been-called-matching` is assumed to be a pattern (when it is a list) we could do more with less verbose syntax.

I agree that `pcase` has an ... interesting pattern language, but it is there to use which is a plus.

Fuco1:
Well, you wouldn't say `'predfunc` you would say `(spy-arg-matcher #'predfunc)`.  That's why I wrap the arguments in the `spy-arg-matcher` wrapper.  If you pass anything which is not this structure, it is taken as a value to `equal` with.  Otherwise the matcher's predicate is called on that position's argument.  The mockery docs explain this in more detail, it's basically the same API.

PR #237, opened 2023-03-21
Do not create trampolines on Emacs 28

### LINKED ISSUE #236: spy-on test fails with native-compilation
I believe this is #230 again. When building with Guix and using native compilation, the tests fail for `spy-on`. Here is the relevant part of the output:
```
The Spy
  `spy-on' function
    replaces a symbol's function slot (0.05ms)
    restores the old value after a spec run (0.05ms)
    allows a spied-on command to be executed as a command (0.09ms)
    can spy on autoloaded functions (0.82ms)
    can spy on non-existing functions (0.03ms)
    only accepts ARG for keywords that use it (0.16ms)
    works on native-compilation primitives  Expected `(spy-on 'buffer-file-name)' not to throw a child signal of `error', but it threw (native-lisp-load-failed "file does not exists" "/homeless-shelter/.emacs.d/eln-cache/28.2-bf9ead57/subr--trampoline-6275666665722d66696c652d6e616d65_buffer_file_name_0.eln") (142.85ms)
    will signal en error if
      used in before-all (0.08ms)
      used directly in describe (0.06ms)
  :to-have-been-called matcher
    returns false if the spy was not called (0.08ms)
    returns true if the spy was called at all (0.07ms)
  :to-have-been-called-with matcher
    returns false if the spy was not called at all (0.10ms)
    returns false if the spy was called with different arguments (0.10ms)
    returns true if the spy was called with those arguments (0.18ms)
  :to-have-been-called-times matcher
    returns error if the spy was called less than expected (0.10ms)
    returns error if the spy was called more than expected (0.09ms)
    returns true if the spy was called the expected number of times (0.08ms)
    use plural words in error message (0.09ms)
    use singular expected word in error message (0.11ms)
    use singular actual word in error message (0.09ms)
  :and-call-through keyword functionality
    tracks calls to the function (0.04ms)
    passes the arguments to the original function (0.07ms)
  :and-return-value keyword functionality
    tracks calls to the function (0.03ms)
    returns the specified value (0.07ms)
    works with strings (0.07ms)
    works with vectors (0.07ms)
    works with symbols (0.07ms)
    works with conses (0.07ms)
    works with lists (11.05ms)
    works with alists (0.06ms)
  :and-call-fake keyword functionality
    tracks calls to the function (0.02ms)
    returns the specified value (0.05ms)
  :and-throw-error keyword functionality
    throws an error when called (0.06ms)
  error-recording functionality
    records the function as called even if it throws an error (0.09ms)
    counts both successful calls and calls that threw errors (0.09ms)
    records args to the function whether it throw an error or not (0.12ms)
    records the signal thrown by a call to the function (0.12ms)
```
and backtrace of the guix build:

```
The Spy `spy-on' function works on native-compilation primitives

  buttercup-fail("%s" "Expected `(spy-on 'buffer-file-name)' not to throw a ...
  signal(buttercup-failed "Expected `(spy-on 'buffer-file-name)' not to thro...
FAILED: Expected `(spy-on 'buffer-file-name)' not to throw a child signal of `error', but it threw (native-lisp-load-failed "file does not exists" "/homeless-shelter/.emacs.d/eln-cache/28.2-bf9ead57/subr--trampoline-6275666665722d66696c652d6e616d65_buffer_file_name_0.eln")

Ran 282 specs, 1 failed, in 314.17ms.
buttercup-run failed: ""
make: *** [Makefile:12: check-buttercup] Error 255
```

This is for v1.30, not sure what is wrong since this was supposed to be fixed. Note that the `homeless-shelter` is noting that `HOME` is not set during the build/test (though I did try setting it to `/tmp` in the package build, perhaps a red herring here as it normally indicates something is relying on `HOME`).

Could be a Guix related issue with how the build is done, but I don't see what that is so far...

snogge:
Bind both `comp-enable-subr-trampolines` and `native-comp-enable-subr-trampolines` and silence the warnings. 
@podiki , can you check whether this fixes #236? I'm not able to recreate on Emacs 28 which I find slightly worrying...

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1284,18 +1284,23 @@ Should always be set to a value that is not `listp', except while
 in a `buttercup-with-cleanup' environment.")
 
 (defvar native-comp-enable-subr-trampolines)
+(defvar comp-enable-subr-trampolines)
 
 (defmacro buttercup-with-cleanup (&rest body)
   "Execute BODY, cleaning spys and the rest afterwards."
-  `(let ((buttercup--cleanup-functions nil)
-         ;; Redefining certain primitive's trampolines will cause problems,
-         ;; see https://github.com/jorgenschaefer/emacs-buttercup/issues/230 and
-         ;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=61880
-         (native-comp-enable-subr-trampolines nil))
-     (unwind-protect (progn ,@body)
-       (dolist (fun buttercup--cleanup-functions)
-         (ignore-errors
-           (funcall fun))))))
+  `(,@(if (fboundp 'with-suppressed-warnings)
+          '(with-suppressed-warnings ((obsolete comp-enable-subr-trampolines)))
+        '(progn))
+     (let ((buttercup--cleanup-functions nil)
+           ;; Redefining certain primitive's trampolines will cause problems,
+           ;; see https://github.com/jorgenschaefer/emacs-buttercup/issues/230 and
+           ;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=61880
+           (comp-enable-subr-trampolines nil)
+           (native-comp-enable-subr-trampolines nil))
+       (unwind-protect (progn ,@body)
+         (dolist (fun buttercup--cleanup-functions)
+           (ignore-errors
+             (funcall fun)))))))
 
 (defun buttercup--add-cleanup (function)
   "Register FUNCTION for cleanup in `buttercup-with-cleanup'."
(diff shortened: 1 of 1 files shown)


podiki:
Hi @snogge, sorry for the delay. Yes that is essentially the same as the fix we used in Guix: https://git.savannah.gnu.org/cgit/guix.git/commit/?id=29be505fceb63d53d1f47312c913e682d19a03d7 (pointed out by someone else, for some reason I didn't think to just cover current and upcoming Emacs versions at the same time).

I checked this PR also passes the tests for me.

As for why you weren't able to recreate the error, it is occurring when the tests are run after native compiling buttercup. In Guix-land, that means building as `guix build emacs-buttercup --with-input=emacs-minmal=emacs` meaning to use full Emacs to do native compilation at build time (and then run the test suite i.e. `make test`). Not sure if that is what you were testing, I haven't tried doing this in Emacs itself directly.

Issue #238, opened 2023-09-27
New failures with native-compilation on emacs-next-minimal-30.0.50-1.9d27b95


nicolas-graves:
```
`pretty' should pretty-print frames  Expected `(buttercup-output)' with value "========================================
suite bc-bt-backtrace

Traceback (most recent call last):
λ (bc-bt-foo
   \"a string that will be truncated in backtrace crop, at least 70 chars long\")
λ (bc-bt-bar
   \"a string that will be truncated in backtrace crop, at least 70 chars long\")
λ (bc-bt-baz
   \"a string that will be truncated in backtrace crop, at least 70 chars long\")
M (or (number-or-marker-p a)
      (signal 'wrong-type-argument (list 'number-or-marker-p a)))
λ (signal wrong-type-argument
        (number-or-marker-p
         \"a string that will be truncated in backtrace crop, at least 70 chars long\"))
error: (wrong-type-argument number-or-marker-p \"a string that will be truncated in backtrace crop, at least 70 chars long\")

" to match the regexp "\\(?:\\`=\\{40\\}
suite bc-bt-backtrace

Traceback (most recent call last):
λ (bc-bt-foo \"\\(?:a string that will be truncated in backtrace crop, at least 70 chars long\\)\")
λ (bc-bt-bar \"\\(?:a string that will be truncated in backtrace crop, at least 70 chars long\\)\")
λ (bc-bt-baz \"\\(?:a string that will be truncated in backtrace crop, at least 70 chars long\\)\")
\\(?:[Mλ] (.*\\(?:
   .*\\)*?.*)
\\)*error: ([^z-a]*)

\\'\\)", but instead it was "========================================
suite bc-bt-backtrace

Traceback (most recent call last):
λ (bc-bt-foo
   \"a string that will be truncated in backtrace crop, at least 70 chars long\")
λ (bc-bt-bar
   \"a string that will be truncated in backtrace crop, at least 70 chars long\")
λ (bc-bt-baz
   \"a string that will be truncated in backtrace crop, at least 70 chars long\")
M (or (number-or-marker-p a)
      (signal 'wrong-type-argument (list 'number-or-marker-p a)))
λ (signal wrong-type-argument
        (number-or-marker-p
         \"a string that will be truncated in backtrace crop, at least 70 chars long\"))
error: (wrong-type-argument number-or-marker-p \"a string that will be truncated in backtrace crop, at least 70 chars long\")

"
```

[...]

PR #239, opened 2023-10-01
test: Change `pretty' test regexp to pass tests on emacs 30.

nicolas-graves:
This is a regexp change that should match test native compilation on both emacs 29 and emacs 30. Should fix https://github.com/jorgenschaefer/emacs-buttercup/issues/238.

### DIFF
--- a/tests/test-buttercup.el
+++ b/tests/test-buttercup.el
@@ -2009,9 +2009,15 @@ text properties using `ansi-color-apply'."
                   "suite bc-bt-backtrace\n"
                   "\n"
                   "Traceback (most recent call last):\n"
-                  "λ (bc-bt-foo \"" (regex ,long-string) "\")\n"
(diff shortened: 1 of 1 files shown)


mekeor:
As discussed in Guix' bug tracker, this PR does not fix the problem: http://issues.guix.gnu.org/66298

mekeor:
Looking at the code of this PR, and my previous comment in the buttercup-bug-report, it's obvious that the new regular expression

```
(seq
 "λ (bc-bt-foo \""
```

won't match

```
λ (bc-bt-foo
```

because of the trailing whitespace. Instead, it should begin like this, I guess:

```
(seq
 "λ (bc-bt-foo"
```

snogge:
I did not notice this PR or #238 before fixing the issue in dfef8cf9c8404567c247b292a4c5752b47e2e4ec.

PR #240, opened 2023-12-03
Use advice-add instead of obsolete defadvice

tarsius:
... and other minor copy editing.

### DIFF
--- a/buttercup-compat.el
+++ b/buttercup-compat.el
@@ -29,8 +29,8 @@
 
 ;;; Code:
 
-;;;;;;;;;;;;;;;;;;;;;
-;; Introduced in 24.4
+;;;;;;;;;;;;;;;;;;;;;;
+;;; Introduced in 24.4
 
 (when (not (fboundp 'define-error))
   (defun define-error (name message &optional parent)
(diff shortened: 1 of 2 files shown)


snogge (approved):

snogge:
Maybe it's time to drop the 24.3 support...

tarsius:
I didn't realize there still was/is support for a pre-nadvice Emacs release. Personally I would drop support for 24 at this point, but that is of course your call. I've actually dropped support for 25 for some packages recently and so far nobody complained (which wasn't the case when I last tried that two years ago, or so.)

snogge:
I try to keep support for as long as possible as buttercup is the kind
of tool that is used to verify that packages work on older Emacsen. 
But I've been looking for excuses to drop support for the older
versions, so I'll drop Emacs 24.3 now.

snogge:
Thank you for the contribution.

tarsius:
> But I've been looking for excuses to drop support for the older versions, so I'll drop Emacs 24.3 now.

Some maintainers of packages that use buttercup will feel the same way. :wink:

Issue #220, opened 2022-06-26
Trace formatting takes ages (with mutual refs?)


jscheid:
tl;dr backtrace formatting took around ~40 seconds for 10 failed tests and then didn't output any stack.

It's more or less instant and prints a useful stack with this:

```el
(defun buttercup--format-stack-frame--print-limits (orig-fun &rest args)
  "Limit `print-level' and `print-length' in ORIG-FUN, forward ARGS."
  (let ((print-level eval-expression-print-level)
        (print-length eval-expression-print-length))
    (apply orig-fun args)))
(advice-add 'buttercup--format-stack-frame
            :around #'buttercup--format-stack-frame--print-limits)
```

My best guess is that what happens is the following:

I have mutual references on the stack, something like:

```el
(cl-defstruct a b)
(cl-defstruct b a)
(let ((a (make-a))
      (b (make-b)))
  (setf (a-b a) b)
  (setf (b-a b) a)
  (func-that-errors a))
```

Because `print-level` and `print-length` are nil by default (at least when running in [makem](https://github.com/alphapapa/makem.sh)), `prin1-to-string` goes into infinite recursion probably until `max-specpdl-size` is exhausted.

I've solved this problem for myself but I had to spend an hour or so debugging it. I recommend making a similar setting the default.

snogge:
Have you tried setting `print-circle` to `t` as described by @alphapapa in https://github.com/jorgenschaefer/emacs-buttercup/issues/218#issuecomment-1151658577 ?

jscheid:
@snogge good idea, I forgot about `print-circle`, thank you.
It seems like it should be exactly what is needed in my case, however I've tried just now and for some reason it doesn't help as much as `print-level` and `print-length` as far as wall time performance goes. In fact is seems to add about 1s to the formatting time, not sure why.

alphapapa:
Yes, if the recursive structures are large enough, the level and length cutoffs could have a larger effect than `print-circle`.

It might be good for Buttercup to bind `print-circle` non-nil by default, anyway.

> In fact is seems to add about 1s to the formatting time, not sure why.

IIUC `print-circle` uses a hash table to track which items have been seen already, which can be some additional overhead, I guess (though whether that overhead is more than that of printing an item depends on the item, I suppose).

snogge:
Fixed by #249

PR #249, opened 2024-08-12
Improve backtrace looks and performance

### LINKED ISSUE #247: Inexplicable hang on Emacs 30 snapshot
Hi,

I'm seeing that my tests for org-ql with Buttercup are hanging exclusively on Emacs snapshot (currently an Emacs 30 pre-release version).  I don't know if this indicates a bug in Buttercup, or an incompatible change in Emacs 30, but I hope that we can find the problem and solve it before Emacs 30.1 is released.  These hangs are happening with the latest version of Buttercup installed automatically from MELPA during the CI run.

Perhaps notably, perhaps related, Guix has disabled tests for org-ql due to what seems to be a similar problem: https://git.savannah.gnu.org/cgit/guix.git/commit/gnu/packages/emacs-xyz.scm?id=3add97c7761e6c58a1d7405f417a49dda5f0a742

```
* gnu/packages/emacs-xyz.scm (emacs-org-ql): Update to 0.8.6.
[arguments]<#:tests?>: Disable tests, which freeze starting with recent
Buttercup releases.
```

I can't reproduce the problem locally, being on a different Emacs version.  But besides that, earlier today while debugging similar issues, I had similar hangs during a Buttercup run on other Emacs/Org/Buttercup versions, and I don't know how to "look inside" a Buttercup run to find out what's hanging.

Any suggestions would be appreciated.  Thanks.

snogge:
Make the backtrace prints better - less unwanted output - with a lot better performance.
Fixes #247 and #220.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -83,8 +83,8 @@ For Emacs < 29:
 The function MUST be byte-compiled or have one of the following
 forms:
 
-\(closure (ENVLIST) () (quote EXPR) (buttercup--mark-stackframe) EXPANDED)
-\(lambda () (quote EXPR) (buttercup--mark-stackframe) EXPR)
+\(closure (ENVLIST) () (quote EXPR) EXPANDED)
+\(lambda () (quote EXPR) EXPR)
 
 and the return value will be EXPR, unevaluated. The quoted EXPR
 is useful if EXPR is a macro call, in which case the `quote'
@@ -98,7 +98,7 @@ ensures access to the un-expanded form."
     ;; * the stackframe marker
     ;; * the macroexpanded original expression
     (`(closure ,(pred listp) nil
-        (quote ,expr) (buttercup--mark-stackframe) ,_expanded)
+        (quote ,expr) ,_expanded)
      expr)
     ;; This a when FUN has not been evaluated.
     ;; Why does that happen?
@@ -107,7 +107,7 @@ ensures access to the un-expanded form."
     ;; * the stackframe marker
     ;; * the expanded expression
     (`(lambda nil
-        (quote ,expr) (buttercup--mark-stackframe) ,_expanded)
+        (quote ,expr) ,_expanded)
      expr)
     ;; This is when FUN has been byte compiled, as when the entire
     ;; test file has been byte compiled. Check that it has an empty
@@ -188,11 +188,9 @@ Does not have the IGNORE-MISSING and SPLIT parameters."
   "Wrap EXPR in a `buttercup--thunk' to be used by `buttercup-expect'."
   (if (fboundp 'oclosure-lambda)        ;Emacs≥29
       `(oclosure-lambda (buttercup--thunk (expr ',expr)) ()
-         (buttercup--mark-stackframe)
          ,expr)
     `(lambda ()
        (quote ,expr)
-       (buttercup--mark-stackframe)
        ,expr)))
 
 (defmacro expect (arg &optional matcher &rest args)
@@ -1015,7 +1013,6 @@ most probably including one or more calls to `expect'."
       `(buttercup-it ,description
          (lambda ()
            (buttercup-with-converted-ert-signals
-             (buttercup--mark-stackframe)
              ,@body)))
     `(buttercup-xit ,description)))
 
@@ -2110,9 +2107,6 @@ ARGS according to `debugger'."
                  (cl-case signal-type
                    ((buttercup-pending buttercup-failed))
                    (otherwise (buttercup--backtrace)))))))
-
-(defalias 'buttercup--mark-stackframe #'ignore
-  "Marker to find where the backtrace start.")
 
 (defun buttercup--backtrace ()
   "Create a backtrace, a list of frames returned from `backtrace-frame'."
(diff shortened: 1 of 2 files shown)


snogge:
@alphapapa @doublep @jcs090218 
I did some major surgery on the backtrace code which really help with performance. 
I'll give you a few days to check it out before I merge. Let me know if I broke anything.

alphapapa:
Hi Ola,

Thanks for the heads-up.  I just did a quick test on the `org-ql` repo, both with all tests passing and with 104 out of 437 specs failing, and performance was excellent: instead of taking minutes to finish outputting the results and the backtraces, it happened in a matter of seconds.  This is a great improvement.  Thank you for working on it!

jcs090218:
Thank you for notifying me. I've done a quick test, and it is working fine. :)

I've opened up a PR in #250 to test downstream Eask and see if it's working. This could make you more confident about the changes. 🤔

snogge:
@doublep , how can I recreate that?

doublep:
~/git/buttercup$ git diff
    diff --git a/tests/test-buttercup.el b/tests/test-buttercup.el
    index 5172c46..02420b3 100644
    --- a/tests/test-buttercup.el
    +++ b/tests/test-buttercup.el
    @@ -172,7 +172,7 @@ before it's processed by other functions."
       (describe "should handle"
         (it "expressions wrapped by buttercup--wrap-expr"
           (expect (buttercup--enclosed-expr (buttercup--wrap-expr-and-eval '(ignore)))
    -              :to-equal '(ignore)))
    +              :to-equal '(INTENTIONALLY-BROKEN-TEST)))
         (it "a closure with expression copy?"
           ;; This is for before Oclosures were added, and is not testable
           ;; once interpreted-function types were added in Emacs 30.
    ~/git/buttercup$ emacs --version
    GNU Emacs 29.4.50
    Development version c0ef9e766eb1 on emacs-29 branch; build date 2024-06-28.
    Copyright (C) 2024 Free Software Foundation, Inc.
    GNU Emacs comes with ABSOLUTELY NO WARRANTY.
    You may redistribute copies of GNU Emacs
    under the terms of the GNU General Public License.
    For more information about these matters, see the file named COPYING.
    ~/git/buttercup$ make test
    ./bin/buttercup -L . tests 
    [...]

      normal-top-level()
      command-line()
      command-line-1(("-L" "." "-l" "buttercup" "-f" "buttercup-run-discover" "-...
      buttercup-run-discover()
      buttercup-run()
      buttercup--run-suites((#s(buttercup-suite "The buttercup--enclosed-expr fu...
      mapc(buttercup--run-suite (#s(buttercup-suite "The buttercup--enclosed-exp...
      buttercup--run-suite(#s(buttercup-suite "The buttercup--enclosed-expr func...
      buttercup--run-suite(#s(buttercup-suite "should handle" #s(buttercup-suite...
      buttercup--run-spec(#s(buttercup-spec "expressions wrapped by buttercup--w...
      buttercup--update-with-funcall(#s(buttercup-spec "expressions wrapped by b...
      apply(buttercup--funcall (closure (t) nil (condition-case err (progn (butt...
      buttercup--funcall((closure (t) nil (condition-case err (progn (buttercup-...
      apply((closure (t) nil (condition-case err (progn (buttercup-expect (let (...
      (closure (t) nil (condition-case err (progn (buttercup-expect (let ((temp0...
      (condition-case err (progn (buttercup-expect (let ((temp0 '(buttercup--enc...
      (progn (buttercup-expect (let ((temp0 '(buttercup--enclosed-expr (buttercu...
      buttercup-expect((closure ((:type . buttercup--thunk) (expr buttercup--enc...
      buttercup-fail("%s" "Expected `(buttercup--enclosed-expr (buttercup--wrap-...
      signal(buttercup-failed "Expected `(buttercup--enclosed-expr (buttercup--w...
    FAILED: Expected `(buttercup--enclosed-expr (buttercup--wrap-expr-and-eval '(ignore)))' to be `equal' to `(INTENTIONALLY-BROKEN-TEST)', but instead it was `(ignore)' which does not match because: (list-elt 0 (different-atoms ignore INTENTIONALLY-BROKEN-TEST)).

    Ran 283 out of 284 specs, 1 failed, in 138.10ms.
    buttercup-run failed: ""
    make: *** [Makefile:12: check-buttercup] Error 255

snogge:
Hmm, but that specific test is not nested.  I'll have to look further.

doublep:
> If you can recreate it with another package I would really like to know about it.

I originally noticed it with `test/project-e` in Eldev's test suite. I did run Buttercup inside Eldev, so initially I suspected that something got broken in my integration code because of these changes. Then I retested it without Eldev at all, on Buttercup directly, and saw the same problem. So, I'm pretty sure that it's a problem with this branch and most likely it's not about nesting level, because it seems to happen both in Buttercup and in unrelated projects.

snogge:
I think I'm on to something now.
The new code works great for unexpected errors in code called from *within* an `except` statement.  That's why it works for the failing `org-ql` tests - there was a bug in `org-ql` that caused `invalid-function` errors. 
Code outside `except` is not handled well on this branch.
The problem with the failing test in buttercup is actually different. That is just a test-mismatch and should really not collect any backtrace at all. What's a meaningful backtrace for `(expect 1 :to-equal 0)`?
There are a few other corner cases that have to be handled as well, like errors in the matcher or in before/after blocks.

snogge:
Newer and even more improved. 
Now also with tests.
@doublep, the problems you reported should be fixed now.
@jcs090218 , the tests #250 passes on this branch.

snogge:
No comments for more than a week, merging this now.

Issue #127, opened 2018-07-23
`:var` does not seem to work with dynamic vars


Fuco1:
I was trying to do something like this: for each spec set a dynamic variable to a test mock value.  I can't use `before-each` and `setq` because that would modify the global state forever.

However, emacs allows `let` binding a dynamic variable and that should then work for all the code called in that scope (dynamic).

```elisp
;; -*- lexical-binding: t -*-

(defvar my-dynamic-var 123)

(describe "Test dynamic vars by binding in describe-form"
  :var ((my-dynamic-var nil))

  (it "does rebind dynamic vars"
    (expect my-dynamic-var :to-be nil)))

(describe "Test dynamic vars by binding in it-form"

  (it "does rebind dynamic vars"
    (let ((my-dynamic-var nil))
      (expect my-dynamic-var :to-be nil))))
```

I would therefore expect this to work, but the test fails.  I think the `it` form is called in a different "dynamic" environment than the describe.  I'm not sure if this can be somehow reconsiled.  A "work around" is to `let` bind the variable in the `it` forms but this leads to major duplication (I have 20-40 `it` forms for each `describe`)

Maybe there could be some other mechanism to do this, like an `unwind-protect` that would set and re-set the variables but this would need a different keyword or notation.

----

There's also the useful idiom to preserve a value of a special variable, for example `kill-ring` like so

```elisp
(let ((kill-ring kill-ring)) ; bind var to itself
  (whatever-here))
```

The `whatever-here` code can do whatever but the global state of the `kill-ring` will be preserved when it returns.  This is quite useful in tests also.  This would also be possible if `:var` worked like this, i.e. preserved the "call stack" and dynamic lookup.

Fuco1:
I think the analysis is as follows: since buttercup relies on lexical scoping, the `let` form generated by `:var` is placed as parent to all the "gathering functions" (`buttercup-before-each`, `buttercup-it` etc) which generate lambdas which capture these variables in their context.  Then this works fine because it's the same variable in all the closures and so before-each affects the it forms as expected.

But then the functions are simply run in a loop from somewhere else and the dynamic scope is no longer extended there.  Yea, not quite sure what to do here, but a feature to modify dynamic scope would be very very rad.

DarwinAwardWinner:
In the meantime, it might be prudent to have `:var` throw an error if lexical binding is not enabled.

snogge:
[...]

PR #255, opened 2025-07-27
Warn when :var is used with a special variable

DarwinAwardWinner:
Buttercup's ":var" and ":var*" forms only support lexical variables, so using a special variable in these forms will almost certainly not do what the test author expects.

See #127.

Notably, `test/test-buttercup.el` now actually issues 2 warnings of this type for the variable `buttercup-reporter-batch--start-time`. I'm unsure if this is intentional, because the comments at those tests suggest something non-standard is happening due to the recursiveness of buttercup testing itself. However, I suspect it is not intentional, and that changes to `buttercup-reporter-batch--start-time` made by those tests are not being properly "quarantined".

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -964,9 +964,9 @@ mainly calls to `describe', `it' and `before-each'."
        when (special-variable-p var)
        do (display-warning
            'buttercup-describe
-           (concat "Possible erroneous use of special variable `"
-                   (symbol-name var)
-                   "' in :var(*) form")))
+           (format
+            "Possible erroneous use of special variable `%s' in :var(*) form"
+            var)))
       ;; Wrap new body in the appropriate let form
       (setq body
             `((,let-form-to-use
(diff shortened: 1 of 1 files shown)


buttercup.el
```diff
@@ -951,19 +951,33 @@ mainly calls to `describe', `it' and `before-each'."
...
+          (let-bindings (elt body 1))
+          (let-body (cddr body)))
+      ;; Verify that all the specified variables are lexical
+      (cl-loop
+       for binding in let-bindings
+       for var = (if (symbolp binding) binding (car binding))
+       when (special-variable-p var)
+       do (display-warning
+           'buttercup-describe
+           (concat "Possible erroneous use of special variable `"
```
snogge, line 967: I'd prefer `(format "... %s ..." var)`.

snogge (commented):
Looks good - one nitpick issue - but I'll have to look into those test warnings.

buttercup.el
DarwinAwardWinner: Fixed. Let me know. If you want me to squash the commits.

snogge:
Thank you.

PR #257, opened 2026-03-09
Scope native-comp trampoline disabling to spy-on only

daym:
`buttercup-with-cleanup` disabled `native-comp-enable-subr-trampolines` for the entire spec execution.  This was a workaround for [Emacs bug#61880](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=61880), where trampoline compilation fails when `spy-on` redefines primitives that the trampoline compiler itself depends on (e.g. `file-exists-p`).

However, disabling trampolines globally has the side effect that `fset` overrides of C primitives become invisible to natively compiled code. This breaks `shut-up`'s message override when called indirectly through native-compiled functions like `display-message-or-buffer` (from `simple.eln`), because without a trampoline, the native code calls the C primitive directly, bypassing the `fset`'d function.

Move the trampoline disabling from `buttercup-with-cleanup` into `buttercup--spy-on-and-call-replacement`, scoping it to just the `fset` call that installs the spy.  This avoids the trampoline compilation issue while allowing other `fset` overrides (e.g. from `shut-up` or `cl-letf`) to work correctly with native-compiled callers.

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1293,14 +1293,33 @@ responsibility to ensure ARG is a command."
       (error "Spies can only be created in `before-each' or `it'"))))
 
 
+(defvar native-comp-enable-subr-trampolines)
+(defvar comp-enable-subr-trampolines)
+
+(defmacro buttercup--without-subr-trampolines (&rest body)
+  "Execute BODY with subr trampoline generation disabled.
+Redefining certain primitive's trampolines will cause problems,
+see https://github.com/jorgenschaefer/emacs-buttercup/issues/230 and
+https://debbugs.gnu.org/cgi/bugreport.cgi?bug=61880"
+  (declare (indent 0) (debug t))
+  `(,@(if (fboundp 'with-suppressed-warnings)
+          '(with-suppressed-warnings ((obsolete comp-enable-subr-trampolines)))
+        '(progn))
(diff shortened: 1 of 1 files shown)


daym:
See also https://codeberg.org/guix/guix/issues/6927

buttercup.el
```diff
@@ -1293,12 +1293,23 @@ responsibility to ensure ARG is a command."
...
 
 
+(defvar native-comp-enable-subr-trampolines)
+(defvar comp-enable-subr-trampolines)
+
 (defun buttercup--spy-on-and-call-replacement (spy fun)
   "Replace the function in symbol SPY with a spy calling FUN."
   (let ((orig-function (and (fboundp spy) (symbol-function spy))))
     (when (buttercup--add-cleanup (lambda ()
                                   (fset spy orig-function)))
```
snogge: Does this `fset` also need to have trampolines disabled?

snogge (commented):
Thank you, this looks good. 
I just have a single worry, that is the `fset` in the cleanup function that restores the function when the spy is disabled.  Does that also need trampolines disabled?

buttercup.el
daym: ... oops.

Good catch!

Updated.

snogge:
Thank you, looks good.

PR #258, opened 2026-03-26
Fix before-all failure/pending not propagating to child specs

### LINKED ISSUE #179: Errors in `before-all` are silently ignored
Consider this test:

```elisp
;;; -*- lexical-binding: t; -*-

(describe
 "A suite"
 (before-all
  (message "well")
  (setq f (1+ f))
  (message "this is unexpected"))

 (it
  "contains a spec with an expectation"
  (expect t :to-be t)))
```

The output is this:
```bash
~/tmp $ emacs -batch -f package-initialize -L . -f buttercup-run-discover
Running 1 specs.

A suite
well
  contains a spec with an expectation (7.30ms)

Ran 1 specs, 0 failed, in 11.24ms.
```

That is unexpected. I would expect that that buttercup should catch all errors in before and after handlers, and print that they fail, and also skip the suite if it fails in a before. 

After all, that is what Mocha does:

```js
describe("A test", () => {
  before(() => {
    throw "Wow, I'm evil";
  });

  it("works", () => {});
});
```

```bash
~/tmp $ mocha test.js

  A test
    1) "before all" hook for "works"


  0 passing (6ms)
  1 failing

  1) A test
       "before all" hook for "works":
     Error: the string "Wow, I'm evil" was thrown, throw an Error :)
      at processImmediate (internal/timers.js:456:21)
```

bbatsov:
When a `before-all` function signals `buttercup-pending` or raises an error, the suite is correctly marked but child specs still run normally, overwriting their status to `passed`. This means `before-all` guards like:

```elisp
(describe "my tests"
  (before-all
    (unless some-condition
      (signal 'buttercup-pending "not available")))
  (it "does something" ...))
```

don't actually skip the specs — they run and potentially fail with unrelated errors.

This PR adds a check after `before-all` completes: if the suite is `pending` or `failed`, the status is propagated to all child specs and nested suites, and the children are skipped. `after-all` functions still run for cleanup.

Fixes #179

### DIFF
--- a/buttercup.el
+++ b/buttercup.el
@@ -1701,16 +1701,38 @@ Do not change the global value.")
     (funcall buttercup-reporter 'suite-started suite)
     (dolist (f (buttercup-suite-before-all suite))
       (buttercup--update-with-funcall suite f))
-    (dolist (sub (buttercup-suite-children suite))
-      (cond
-       ((buttercup-suite-p sub)
-        (buttercup--run-suite sub))
-       ((buttercup-spec-p sub)
-        (buttercup--run-spec sub))))
+    ;; If before-all marked the suite as pending or failed,
+    ;; propagate to all children and skip running them.
+    (if (memq (buttercup-suite-or-spec-status suite) '(pending failed))
+        (buttercup--propagate-suite-status suite)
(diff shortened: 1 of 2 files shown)


snogge:
Thank you.

